PR #832, opened 2017-06-12
Issue #793

### LINKED ISSUE #793: Sequential editing of entities in EGI (simple masters only)
### Description

The only way currently to edit entities is through entity masters. It was a deliberate decision not to make grids such as EGI editable. There are numerous good reasons for that, one of those being UX consistency.

However, there are cases that require for all or at least the majority of entities that are displayed in EGI to be edited. Currently users are forced to tap "edit" on each individual EGI row one by one to open a corresponding entity master for editing, perform editing and then save change. This is too laborious.

Need to provide a generic action that would kick-in the process of sequential editing of selected entities from the current EGI page (if no entities are selected then assume editing of all entities as if all are selected). Such action needs to be tightly integrated with EGI and "know" the sequence of entities in EGI. 

Once the action is activated, a master for the first selected entity should open. The following rules apply:

1. Pressing button CANCEL should close the current master and abort the whole editing process.
2. Once an entity was edited, pressing button SAVE should perform the standard save action and if successful load the next entity from the list of selected entities. If there are no unedited entities in the list then successful saving of the current one should simply close the dialog as per usual.
3. There could be cases where the user might decide not to edit the currently open entity. In this case button SAVE would be disabled, providing no way to move to the next entity. In order to remedy such cases, the entity master dialog (`tg-custom-action-dialog`) should have button "next" (`av:skip-next`) instead of "close" in its title bar. Tapping this button should skip the current entity (if not changed) and load the next one from EGI.

This issue is reminiscent of #285 for continuous entity creation.

## Expected outcome

Significantly improved UX for editing multiple entities sequentially.

oleh-maikovych:
resolve #793

### DIFF
--- a/platform-web-resources/src/main/java/ua/com/fielden/platform/web/test/server/config/StandardActions.java
+++ b/platform-web-resources/src/main/java/ua/com/fielden/platform/web/test/server/config/StandardActions.java
@@ -18,6 +18,7 @@ import ua.com.fielden.platform.reflection.TitlesDescsGetter;
 import ua.com.fielden.platform.web.PrefDim;
 import ua.com.fielden.platform.web.action.post.FileSaverPostAction;
 import ua.com.fielden.platform.web.action.pre.ExportPreAction;
+import ua.com.fielden.platform.web.action.pre.SequentialEditPreAction;
 import ua.com.fielden.platform.web.centre.api.actions.EntityActionConfig;
 import ua.com.fielden.platform.web.centre.api.context.IEntityCentreContextSelectorDone;
 import ua.com.fielden.platform.web.minijs.JsCode;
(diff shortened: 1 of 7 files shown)


platform-web-ui/src/main/java/ua/com/fielden/platform/web/action/CentreConfigurationWebUiConfig.java
```diff
@@ -81,8 +88,95 @@ public JsCode build() {
...
+            }
+
+            @Override
+            public EntityActionConfig mkAction(final Class<? extends AbstractEntity<?>> entityType) {
+                return mkAction();
+            }
+
+        },
+
+        SEQUENTIAL_EDIT_ACTION {
```
01es: The locality of this action implementation needs to be improved. For example, currently `EDIT_ACTION` is an application rather then a platform concern. The same should be followed in relation to action `SEQUENTIAL_EDIT_ACTION`.

platform-web-resources/src/main/java/ua/com/fielden/platform/web/test/server/config/StandardActions.java
```diff
@@ -146,6 +148,85 @@ public EntityActionConfig mkAction(final Class<? extends AbstractEntity<?>> enti
...
+                contextConfig = context().withCurrentEntity().withSelectionCrit().withComputation(computation);
+            } else {
+                contextConfig = context().withCurrentEntity().withSelectionCrit().withComputation(entity -> entityType);
+            }
+
+            return action(EntityEditAction.class).withContext(contextConfig.build()).preAction(new IPreAction() {
+
+                @Override
+                public JsCode build() {
+                    return new JsCode("\n"
```
01es: Please take a close look at the `SEQUENTIAL_EDIT_ACTION` implementation in here and in TG Air. Are there common parts that should really be extracted into the core platform functionality? (Hint: there are!)

PR #872, opened 2017-07-14
Issues #327 and #335

### LINKED ISSUE #327: EGI: support for property inclusion/exclusion (customizable columns)
In order to improve EGI responsiveness and usability it is proposed to implement user customisation whereby the user should be able to specify what properties are to be included/excluded for display.

Each Entity Centre has a model behind that specifies what properties should be included into EGI. Users should be able to amend the default model by explicitly specifying what properties out of included by default should be displayed in EGI and which ones should be omitted. This feature should be relatively easy to implement, but would immediately empower users with the ability to adjust the data view by choosing only what they need to see and/or to fit their screen size. 

Property inclusion/exclusion should be invoked via a `icons:filter-list` button in the right block of EGI toolbar actions. This action should invoke a configuration dialog that would list (ref [iron-list](https://elements.polymer-project.org/elements/iron-list?view=demo:demo/selection.html&active=iron-list) all included by default into the Entity Centre model properties and provide a capability to specify which ones should be included or excluded. This dialog should be fitted with "Apply" and "Cancel" buttons that would apply or cancel user changes.

Application of user changes should be immediately applied to the effected EGI (no reloading is required due to a complete data availability at the client side).

The following sketch illustrates the EGI resultset configuration dialog.

screen shot 2015-11-26 at 11 27 46

Please also refer to section "Customizable Columns" in [Design better data tables](https://uxdesign.cc/design-better-data-tables-4ecc99d23356) for inspiration.

**Side note: it is envisaged that the described property inclusion/exclusion should be automatically persisted; also the discussed configuration dialog can be used for specifying sorting order and the order of columns in EGI. There should also be a "quick view" facility for each EGI row to display the values for excluded properties -- such feature might supersede the need to have the EGI card representation... All of this is to be covered by separate issues.**

### LINKED ISSUE #335: EGI: support custom order of properties in a row (customizable columns)
In addition to property inclusion/exclusion (#327) and ad-hot data sorting (#334), the EGI property configuration dialog should support property ordering that would control the order of columns in the grid.

It is desirable that property ordering would be performed by D'n'D of property items in the list. There should be an appropriate D'n'D affordance in a form of a visual cue that would suggest that property items are draggable. For example, Gmail client displays a dotted bar when hovering over an item with a mouse cursor as depicted below. A similar approach could be used, but with the dotted bar always displayed for all list items (i.e. no hovering required, which would make it suitable for touch devices without a mouse.

screen shot 2015-11-26 at 12 20 35

Some additional discussion and inspirations can be found [here](http://ux.stackexchange.com/questions/26121/what-is-a-good-visual-cue-for-drag-and-drop-for-reordering-a-list-of-items). Please also refer to section "Customizable Columns" in [Design better data tables](https://uxdesign.cc/design-better-data-tables-4ecc99d23356).

jhou-pro:
resolve #327 resolve #335


01es:
Additional change is required to improve the mouse cursor behaviour in order to avoid a "flickering effect". Please refer a screencast below.

mouse-cursor-flickering

PR #879, opened 2017-07-20
Issue #878

### LINKED ISSUE #878: It should not be possible to move Entity Master above the top of the page
### Description
Entity Masters can be moved around the page by dragging them by the title bar. This is a very useful feature that allows users to conveniently arrange open Entity Masters to their needs.

There are situations, however, where users may drag an Entity Master so that the title bar moves above the top of the page, "under" the browser address bar. This makes it impossible then to move that Entity Master anywhere else because the title bar is not visible to grab it. Please refer a screenshot below.

screen shot 2017-07-20 at 13 45 04

Although it is possible to close such Entity Masters (with `Alt+x` for compound or `Cancel` for simple masters), users may get disoriented by this situation. Thus, it is require to restrict the ability for an Entity Master to be dragged above the top of the page.

### Expected outcome
Improved UX where users would not "get stuck" with Entity Masters that cannot be moved anywhere.

oleh-maikovych:
Resolve #878

### DIFF
--- a/platform-web-ui/src/main/web/ua/com/fielden/platform/web/egi/tg-custom-action-dialog.html
+++ b/platform-web-ui/src/main/web/ua/com/fielden/platform/web/egi/tg-custom-action-dialog.html
@@ -201,7 +201,13 @@
                 /**List of child dialogs*/
                 _childDialogs: {
                     type: Array
-                }
+                },
+                
+                /**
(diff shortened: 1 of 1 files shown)


01es:
@oleh-maikovych I still managed to move the window "beyond" the browser. Please refer the screencast below.

can-still-move-window-beyound

Issue #892, opened 2017-08-14
Unexpected mutation of previous EQL statements model -- side-effect of successive chaining of EQL stmt operators.


nazchk:
### Description
While building the EQL query statement via progressive interfaces API there are combinations where successive step mutates the state of previously created statement.

Need to eliminate this to ensure consistency of developer experience.

In order to reproduce the problem the following test has been created:

```Java
final ICompoundCondition0<TgVehicle> qryStmt = select(TgVehicle.class).as("veh").where().prop("veh.model").isNull();
final ICompoundCondition0<TgVehicle> equalQryStmt = select(TgVehicle.class).as("veh").where().prop("veh.model").isNull();

assertTrue(qryStmt.equals(equalQryStmt));
		
equalQryStmt.groupBy().prop("veh.model");
	
assertTrue(qryStmt.equals(equalQryStmt));
```

PR #905, opened 2017-09-14
Issue #892

nazchk:
Review PR #892


platform-pojo-bl/src/main/java/ua/com/fielden/platform/entity/query/fluent/AbstractQueryLink.java
```diff
@@ -21,4 +21,29 @@ public String toString() {
 	private void setTokens(final Tokens tokens) {
 		this.tokens = tokens;
 	}
+
+	@Override
+	public int hashCode() {
```
01es: `tokens` are mutable, which makes it unsuitable for computing a hash code... For example, if instance of  `AbstractQueryLink` is added to a hash-set and the `tokens` are assigned or mutated subsequently then the same object would no longer be recognised as "contained" in the set.

PR #908, opened 2017-09-18
Issue #786

### LINKED ISSUE #786: Validation Prototype: avoid new entity to be produced multiple times
### Description
Since the entity is always produced during `Entity Master`'s retrieval action (new entities only), it is understood that such process could be avoided afterwards during validation, saving and autocompletion processes.

Such enhancement could enable restoration of entity during autocompletion process in cases where such restoration requires parent context. Also this could improve entity restoration in overall due to the hypothesis (which needs to be proved) that, most likely, parent context will not be needed at all if entity producer is not used.

This issue should be served as a spike to prove that described enhancement is fully reasonable (no edge-cases should exist that will break current functionality) and could be implemented in congruent timeframe.

### Expected outcome
Reliable and more efficient entity master processes. Always workable autocompletion process.

### Actual outcome 
Not always workable autocompletion process while parent context is needed for master entity restoration.

jhou-pro:
resolve #786


01es:
This PR should be merged only once the corresponding TTGAMS changes are approved.

PR #928, opened 2017-10-27
Issue #924

### LINKED ISSUE #924: Insertion point with pagination buttons
### Description

In some cases it is very useful to navigate (paginate) data from entity centre's insertion point, especially  when insertion point is expanded and there is no way to interact with pagination buttons on EGI. One of such case is `tg-planning` insertion point. Therefore it is needed to provide an ability for one to add pagination buttons to the insertion point and make them visible only when insertion point is expanded.

### Expected outcome

Insertion point can have pagination buttons when it is in expanded mode. This feature should be used for `tg-planning` insertion point.

oleh-maikovych:
resolve #924


01es:
The outline of the circled button touches the top of the panel. At least 2 pixel padding from the top is needed.

screen shot 2017-11-01 at 16 46 20

PR #989, opened 2018-02-15
Issue #539

### LINKED ISSUE #539: Document attachments as files
### Description

There is a need to associate various documents with business entities that are managed by an application. This issue should be concerned only with documents that are represented as files (e.g.PDF or DOC(x) documents). 

However, it should be understood that the concept of a document is richer and may include artefacts from Google Docs, Microsoft SharePoint or some document management system -- all of these are very specific and, although could be abstracted out behind a generic API, would require a very specific implementation and, thus, issues to cover that work.

It is envisaged that file-based document attachments should be concerned strictly with document viewing (i.e. no editing) and should include an internal version tracking. The internal version tracking is required to ensure explicit and transparent handling of situations where one document file gets replaced with a different version of the same file. The system should recognise the fact that a file was changed by [computing its checksum](http://stackoverflow.com/questions/304268/getting-a-files-md5-checksum-in-java) and recording a new version automatically. A clear indication of the document version should be provided to the end application user.

The uploaded files should be stored on a files system in a designated location. The directory structure and file names should be governed by the system internally in order to achieve the needed functionality. However, the original file names should be preserved as part of the meta-data that is captured by the `Attachment` entity.
### Expected outcome

A functionality to manage file-based document attachments that can be associated (attached) with any business domain entity and made easily available for the download or upload in the context of a specific entity (as opposed to just the `Attachment` entity, its centre and master).

As a side note, a full-text search based on Lucene is one of the future platform features, which can/should be expanded to index the attached documents.

01es:
Resolve #539


01es:
This PR should only be merged once issue https://github.com/fieldenms/TTGAMS/issues/641 is completed.

PR #992, opened 2018-02-17
Issue #874

### LINKED ISSUE #874: Generic D’n’D capability
### Description

Need to provide Generic Drag and Drop support for tg application. For that purpose one should implement a `behavior` name it: `TgDragFromBehavior` this behavior should be implemented by components those are interested in exporting data via drag-and-drop process. This behavior should have next methods:

- `getElementToDrag`
- `getDataToExport`

Also this behavior should allow to mark elements in the implementing components. This mark should be used as anchor for user to grab it and drag. Also This mark could be some attribute or event listener for `track` event. This event listener will trigger different phases of drag and drop process.

Another behavior that should be implemented, name it: "TgDropToBehavior" should provide a contract for component to import data, and should have the following methods:

- `preparePlaceForData(elementToDrop, dataToDrop)`
- `importData(dataToDrop)`

 Every element that is interested in D'n'D its data, should implement both of those behaviors.

The following articles might be inspiring or even very helpful during implementation of this issue:

- [Drag and Drop example how it may looks like](https://www.uplabs.com/posts/drag-drop-tasks) - demonstrates
- [setCapture method to handle cursors and events during Drag and drop process](https://developer.mozilla.org/en-US/docs/Web/API/Element/setCapture)
- [polifill for setCapture method if it's not supported in your browser](https://gist.github.com/jnm2/fb211eeaf1c27023fbb76d2868957a68)

oleh-maikovych:
resolve #874


01es:
There is minor concern re misalignment of top-level actions and EGI checkboxes as per image below. However, at this stage there is no clear solution to this problem...

image

platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-drag-example.html
```diff
@@ -0,0 +1,33 @@
...
+            font-family: 'Roboto', 'Helvetica Neue', Helvetica, Arial, sans-serif;
+            font-size: 10pt;
+        }
+    </style>
+
+</head>
+
+<body unresolved class="layout horizontal">
+    <tg-drag-from-component-example style="margin-right: 300px;"></tg-drag-from-component-example>
+    <tg-drop-to-component-example></tg-drop-to-component-example>
```
01es: This example seems to be broken. It should either be fixed or removed if it served its purpose.

platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-drop-to-behavior.html
```diff
@@ -0,0 +1,24 @@
...
+<link rel="import" href="/resources/reflection/tg-polymer-utils.html">
+
+<script>
+    (function () {
+
+        Polymer.TgBehaviors = Polymer.TgBehaviors || {};
+        Polymer.TgBehaviors.TgDropToBehavior = {
+
+            canDropType: function (types, type) {
+                for (var i = 0; i < types.length; ++i) {
```
01es: It's time to start using `let`... or, in this case, the use of [Array.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) would be most appropriate...

platform-web-ui/src/main/web/ua/com/fielden/platform/web/egi/tg-entity-grid-inspector.html
```diff
@@ -788,7 +830,7 @@
                 _lastSelectedIndex: Number
             },
 
-            behaviors: [Polymer.TgBehaviors.TgTooltipBehavior, Polymer.IronResizableBehavior, Polymer.IronA11yKeysBehavior, Polymer.TgBehaviors.TgShortcutProcessingBehavior],
+            behaviors: [Polymer.TgBehaviors.TgTooltipBehavior, Polymer.IronResizableBehavior, Polymer.IronA11yKeysBehavior, Polymer.TgBehaviors.TgShortcutProcessingBehavior, Polymer.TgBehaviors.TgDragFromBehavior],
```
01es: Need to add support for `TgDropToBehavior` in the future.

01es:
This PR should only be merged once https://github.com/fieldenms/TTGAMS/pull/887 is ready to be merged.

PR #1008, opened 2018-03-20
Issue #1007

### LINKED ISSUE #1007: Centre URI enhancement: apply uri parameters only once and remove it from history.
### Description

Need to enhance centre URI with query parameters functionality according to the following requirements:

-  [x] 1. URI parameters of entity centre should be applied only once (i.e. URI should be applied and removed from history being replaced with actual centre URI without parameters.) This would prevent from running this entity centre again with the same parameters after `discard` was pressed.
- [x] 2. When applying centre URI parameters one should clear all mnemonic parameters.

### Expected outcome

Entity centre should work in more predictable way after discard was pressed.

oleh-maikovych:
Resolve #1007

### DIFF
--- a/platform-web-ui/src/main/web/ua/com/fielden/platform/web/centre/tg-entity-centre-behavior.html
+++ b/platform-web-ui/src/main/web/ua/com/fielden/platform/web/centre/tg-entity-centre-behavior.html
@@ -411,12 +411,14 @@
                         self._setQueryParams();
                         if (self.autoRun || self.queryPart) {
                             self.run();
+                            delete self.queryPart;
                         }
                     });
                 } else {
                     self._setQueryParams();
                     if (self.autoRun || self.queryPart) {
                         self.run();
+                        delete self.queryPart;
                     }
                 }
             },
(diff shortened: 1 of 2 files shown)


platform-web-ui/src/main/web/ua/com/fielden/platform/web/centre/tg-entity-centre-behavior.html
```diff
@@ -820,6 +825,7 @@
                         secondEditor.commit();
                     }
                 });
+                delete this.queryPart;
```
01es: This `delete` causes a problem whereby selection criteria values that are assigned by default (values, not mnemonics) get discarded before automatically running the centre.

It has been observed that the only reason for the introduction of this `delete` statement is to remedy the problem with action "Discard" leading to re-running of entity centres that were opened via an URI with parameters. It was originally thought that such "re-run" behaviour was caused by the presence of parameters in the URI upon "Discard".  However, having those parameters now stripped off, but this `delete` statement removed, manifests in the same "unexpected" way -- re-running the centre.

Thus, item 1 of this issue cannot be considered "done". The solution should not discard default selection criteria values and should not re-run the centre upon "Discard".

PR #1018, opened 2018-03-29
Issue #1017

### LINKED ISSUE #1017: Cascading confirmation dialog
### Description

In some cases there is a need to open confirmation dialog more then once during short period of time this may cause conflict between those two invocation of confirmation dialog. Such case requires confirmation dialog enhancement so that it can be opened in cascade (i.e next invocation of confirmation dialog should wait until the previous confirmation dialog will be close.)

### Expected outcome

Simultaneous confirmation dialog invocation should open them in cascade

### Actual outcome 

Next confirmation dialog invocation causes dialog to change it's message and buttons.

oleh-maikovych:
resolve #1017

### DIFF
--- a/platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-confirmation-dialog.html
+++ b/platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-confirmation-dialog.html
@@ -49,10 +49,19 @@
 
             Polymer({
                 is: 'tg-confirmation-dialog',
+                
+                showConfirmationDialog: function (message, buttons) {
+                    if (this._lastPromise) {
+                        this._lastPromise = this._lastPromise.then(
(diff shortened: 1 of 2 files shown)


platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-confirmation-dialog.html
```diff
@@ -49,10 +49,19 @@
 
             Polymer({
                 is: 'tg-confirmation-dialog',
+                
+                showConfirmationDialog: function (message, buttons) {
+                    if (this._lastPromise) {
+                        this._lastPromise = this._lastPromise.then(
+                            value => this._showConfirmationDialog(message, buttons), 
+                            reason => this._showConfirmationDialog(message, buttons));
```
01es: @oleh-maikovych is there any particular reason for providing optional handler `onRejected` in this case?

PR #1104, opened 2018-07-27
Issue #1058

### LINKED ISSUE #1058: UTC Date/Time support: alternative formatting
### Description

Issue #642 introduced support for UTC dates (proper conversion at the server and client ends, etc.).

Based on feedback from Airways, the way UTC dates are used in practice require to have those values displayed in the ISO standard "YYYY-MM-DD". 

This alternative format is required for both displaying and editing.

Please consider value approximation support when implementing this issue.

### Expected outcome

The ability to represent UTC dates in the ISO format for displaying and editing.

oleh-maikovych:
resolve #1058


01es:
@oleh-maikovych Please investigate failing web tests in `tg-datetime-picker-time-zones-for-24hours-for-persisted-entity.html`.

platform-web-ui/src/main/web/ua/com/fielden/platform/web/editors/tg-datetime-picker.html
```diff
@@ -432,6 +436,20 @@
                 }
                 return undefined;
             },
+            
+            /**
+             * Tries UTC formats one by one and returns valid moment or null if there are no appropriate format.
+             */
+            _tryUTCFormats: function (stringValue) {
```
01es: Can we do somethings like the snipped below?
```
const timeZone = this.timeZone;
const tryingMoment = Object.values(timeZoneFormats['UTC'])
                     .map(f => _momentTz(stringValue, f, true, timeZone))
                     .find(m => m.isValid());
return tryingMoment ? tryingMoment : null;
```

PR #1119, opened 2018-08-22
updated polymer to version 1.11.3

oleh-maikovych:


01es:
Overall it looks good, but requires more testing. Some things seem to be a bit slower (e.g. selecting bars on the planning component in TG Air), while others a bit faster (e.g. complex EGI seems to be rendering faster).

PR #1132, opened 2018-09-05
Backlog 107

krvavych:


01es:
Commit dc1f2b3 broke the attachment-as-property functionality while not achieving the original objective. Alternative implementation has been provided as commit fd37c27.

Commit d46eeae was cherry-picked into develop.

Issue #1185, opened 2018-11-26
Bar charts: data grouping (stacked and side-by-side)


oleh-maikovych:
### Description

At the moment bar charts that were developed as part of issue #1031 only support a single bar for a specific category or group. Bar charting capability should be enhanced to support grouping of more than one value by a specific category. Both stacked (like in this example http://bl.ocks.org/KatiRG/5f168b5c884b1f9c36a5) and side-by-side (like in this example https://bl.ocks.org/mbostock/3887051) group representations are required. 

In addition, bar charts should support horizontal lines as a means to specify various threshold values.

The following items need to be implemented:

- [x] The bar chart implementation should be enhanced to support grouped bar charts (stacked and side-by-side).
- [x] Enhance triple decker API to specify grouping value/property and to specify the details view for each value in a group.
- [x] Enhance triple decker API to specify horizontal lines (threshold values).

### Expected outcome

Ability to group values and setup threshold values (as horizontal lines) for individual deckers.

PR #1207, opened 2018-12-10
Issue #1170

### LINKED ISSUE #1170: Entity Master: navigating across entities with compound master to preserve activated menu
### Description

As an enhancement to issue #1055, need to implemented support for preserving the currently activated menu item when navigating across entities that are presented by a Compound Entity Master. 

Presently, navigating away from the current entity results in activating menu "Main" regardless of what menu item was active for the entity navigated from. This needs to be changed to activate that menu items, which was active for the entity that was loaded before moving to another (next, previous, last or first) entity.

### Expected outcome

Ability to view master/detail associations by navigating across entities represented by Compound Entity Masters.

oleh-maikovych:
resolve #1170


01es:
@oleh-maikovych was there any particular reason to base #1170 on #1185?

PR #1208, opened 2018-12-10
Issue #1206

### LINKED ISSUE #1206: Entity Centre: hide selection criteria action (funnel icon) in case of no criteria
### Description
The are many cases where Entity Centres -- embedded, those used for insertion points and details -- have no selection criteria exposed to users. Such centres are always auto-run and have no need for users to switch to the selection criteria panel.

Based on the most recent user feedback, it would be best to remove/hide the "funnel" action from the EGI's top action panel to prevent users from switching to an empty selection criteria panel by mistake.

### Expected outcome
Improved usability of Entity Centres that have no selection criteria by design.

### Actual outcome 
An empty selection criteria panel that confuses users.

oleh-maikovych:
resolve #1206 merge after issue #1170 will be merged


01es:
@oleh-maikovych is there any particular reason #1206 is based on #1170?

PR #1229, opened 2019-01-06
Issue #1145

### LINKED ISSUE #1145: Entity Centre: gradual adaptation of saved configurations during domain evolution
### Description

At this stage there is already a good level of adapting capabilities for saved centre configurations. What we mean here is that when domain entity type change (for e.g. property removed) then new version of server will try its best to adapt existing saved centre configurations during their load so that they will be loaded gracefully (for e.g. with all criteria values except the value for the property that was removed).

There are several places of complexity here: property removed / refactored in domain entity; calculated property changed in Centre DSL configuration; domain entities added / moved / removed.

For now we have several major outstanding items:

1. When changing the order of entity type registration in `ApplicationDomain` (and therefore in serialiser) existing configurations, pertaining to `shifted` types, crash.
2. When changing entity type, for which there exists configuration with at least one crit-only single criteria value of that type, that configuration crashes after server restart.
3. There is an impression when removing property from domain type that corresponding configurations (with property existing in selection criteria) crash. This is due to excessive exception logging.
4. When removing property from domain type corresponding configurations (with property existing in result set) crash. This was somewhat patched with efafe84c38fe59e61fd847850111e754d92e7a4a.

Most of these items are related to the mechanism of serialisation which uses `Kryo` serialiser and serialises whole `CentreDomainTreeManagerAndEnhancer` with some its parts marked as "changed". It is required to replace such mechanism with better one that does not serialise unnecessary information (for example `rootTypes`, which causes item 1.) and that resiliently adapts to domain / Centre DSL config changes.

### Expected outcome
Better level of adaptation of existing centre configurations in accordance to business domain changes.

jhou-pro:
resolve #1145


01es:
This PR should be merged into `develop` in sync with changes in TG Air.

Issue #1236, opened 2019-01-14
Enhance EGI's entity equality logic to use Reflector's one.


oleh-maikovych:
### Description

EGI has own private equals method for entities that relays on the fact that entity has `id` property. But there is a use case when EGI entities don't have `id` property but still there is need to compare such entities based on their keys. In that case EGI equals method should be enhanced to use reflector's `equalsEx` logic.

### Expected outcome

EGI's equals method should use Reflector's `equalsEx` method to compare entities.

### Actual outcome 

EGI fails to compare entities which don't have `id` property.

PR #1240, opened 2019-01-31
Fix for issue #1236

oleh-maikovych:
Introduced fix in equality method that handles the exception when key is not present.


platform-web-ui/src/main/web/ua/com/fielden/platform/web/egi/tg-entity-grid-inspector.html
```diff
@@ -964,7 +964,12 @@
                 if (a && b && a.get('id') && b.get('id')) {
                     return a.get('id') === b.get('id');
                 }
-                return this.$.reflector.equalsEx(a, b);
+                try {
+                    return this.$.reflector.equalsEx(a, b);
+                }
```
01es: Please combine with the next line to hav `catch` following `{`.

PR #1246, opened 2019-02-07
Issue #1245

### LINKED ISSUE #1245: Standard edit action: support navigation through erroneous / inaccessible entities
### Description
At this stage when we navigate from correct entities to incorrect one (by means of exception in corresponding producer) we get empty master with appropriate toast and error message inside master. However further navigation is not possible. Also incorrect number of current entity is sometimes displayed (`NAN`). Broken behaviour can be remedied by closing and opening master again.

It is required to be able to navigate through erroneous masters. This is legal case especially when the user has no rights to open corresponding entity. It is worth noting that such situation is primarily related (but not limited) to polymorphic entities (example fieldenms/TTGAMS#1510 item 5).

### Expected outcome
Navigation through erroneous / inaccessible entities in standard edit facility.

oleh-maikovych:
resolve #1245


platform-web-ui/src/main/java/ua/com/fielden/platform/web/view/master/api/with_master/impl/EntityNavigationMaster.java
```diff
@@ -37,8 +37,13 @@ public EntityNavigationMaster(final Class<EntityNavigationAction> entityType, fi
               + "                     this._prevCurrBindingEntity = e.detail.value;\n"
               + "                 }\n"
               + "             }.bind(self);\n"
+              + "             self._handleBindingEntityRetrievedError = function (e) {\n"
+              + "                 this.fire('tg-master-navigation-error');\n"
+              + "                 this._previousMaster = this.$.loader.loadedElement;\n"
```
01es: Shouldn't this assignment be before firing the event?

platform-web-ui/src/main/web/ua/com/fielden/platform/web/egi/tg-custom-action-dialog.html
```diff
@@ -980,48 +1010,58 @@
...
                 if (e.detail.shouldResetSpinner) {
                     this._resetSpinner();
                 }
             },
             
+            // Inoked by navigation buttons it turns on spinner on navigation button and makes blocking layer visible. Also it resets error message if it is present due to previous navigation action.
             _handleActionNavigationInvoked: function (e) {
+                //Reset dataLoaded property to ensure that blocking layer will work correctly and appropriate event handlers will be invoked.
                 this._dataLoaded = false;
                 if (this._errorMsg) {
```
01es: Is this condition really needed? Could there be just `this._errorMsg = null;` without a condition?

platform-web-ui/src/main/java/ua/com/fielden/platform/web/view/master/api/with_master/impl/EntityNavigationMaster.java
oleh-maikovych: Actually it doesn't meter whether it's before or after. _previousMaster is needed only to identify the fact that dialog will change it's master type and it happens only when during next navigation action execution. The event will be handled only in action's preAction during current navigation action execution

platform-web-ui/src/main/web/ua/com/fielden/platform/web/egi/tg-custom-action-dialog.html
oleh-maikovych: Yes, this assignment can be done without condition, i'll change it.

PR #1253, opened 2019-02-15
Issue #1252

### LINKED ISSUE #1252: Activatable Entity: prevent potentially incorrect refCount calculation based on prevValues
### Description

As part of issue #646 the use of `prevValue()` was introduced to replace unreliable Hibernate `load` call when handling dirty activatable properties. It is suspected that under certain conditions, such as changing an activatable property more than once to different values before saving changes, the use of `prevValue()` could lead to invalid recalculation of `refCount`.

This issue covers the work to determine whether the above is a problem and to develop a correction if it is. 

### Expected outcome
Improved processing of activatable entities.

01es:
Resolve #1252


01es:
Additional tests are in progress...

PR #1337, opened 2019-06-12
Issue #1331

### LINKED ISSUE #1331: Client-side caching of static resources
### Description
There are many different types of resources in TG applications that are fetched during client-side application lifecycle. These resources include Polymer library elements, our elements (`tg-entity.centre.js`, `/centre_ui/fielden.work.ReWorkActivity.js` etc.), iconsets, third-party libs (`/resources/postal/2.0.5/postal.min.js` etc.) and others.

These resources evolve from one release to another. New version of TG-based app can bring new ones, some may be removed and another changed. However, most of them are static until the next release come into life.

There are also non-static resources (requests) that pertain to data modification and retrieval. Entity retrieval / saving and centre running requests are an obvious examples to these.

Currently we have [almost] all static resources bundled into single one *vulcanised* resource to reduce the number of requests when client application loads. But do we really need to fetch the same bundle over and over again when client application reloads? An obvious answer is `no`: we can cache that bundle and reuse it until new app version arrives.

Proper version handling is required when the server application gets new version. User should be warned after reloading of client application that there is a new version. As the first step we could guide the user to close all application tabs and open application again. That will promote new version of the application in its fullness.

As part of this issue we should also test how un-bundled application [deployment mode] will work against HTTP2 protocol in conjunction with static resources caching. Slow networks and first-time loading should be considered when measuring and then careful decision should be made.

### Implementation details
`Service Workers` and `Cache` API.

### Expected outcome
Better performance of client-side application loading with the aim of instantenious non-blocking experience if possible.

jhou-pro:
resolve #1331


01es:
Closing PR until the issue is fully implemented.

PR #1404, opened 2019-08-14
Issue #1388

### LINKED ISSUE #1388: Represent the title of a module menu item in the browser tab
### Description
At the moment the document title for TG applications is static and reflects the application name and version.

Based on the user feedback, there is a need to update the document title dynamically based on a module menu item loaded. This is especially convenient in cases where multiple browser tabs or windows are used.

In most cases module menu items represent Entity Centres, which may have a custom configuration loaded. The document title should have the same value as the one for the centre. Moving away from a module menu item to the main menu should restore the document title to its static value (i.e. application name and version).

image

#### Implementation remark
Document title can be changed by assigning `document.title = ...`.

### Expected outcome
Improved usability in the multi-tab and multi-window configuration.

oleh-maikovych:
resolve #1388

### DIFF
--- a/platform-web-ui/src/main/web/ua/com/fielden/platform/web/app/tg-app-template.js
+++ b/platform-web-ui/src/main/web/ua/com/fielden/platform/web/app/tg-app-template.js
@@ -37,7 +37,7 @@ const template = html`
     <tg-message-panel></tg-message-panel>
     <div class="relative flex">
         <neon-animated-pages id="pages" class="fit" attr-for-selected="name" on-neon-animation-finish="_animationFinished" animate-initial-selection>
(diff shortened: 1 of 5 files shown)


01es:
@oleh-maikovych the functionality required by the issue works very well. However, testing suggested a minor improvement -- a module title (i.e. not only module item title) should also be covered. Please refer the screen capture below.

image

Without this feature there might some confusion in cases where there was some module item selected, and user navigates to another module using F3 feature. Please refer the screen capture below.

image

There is an edge-case scenario here -- when user had a module menu selected and then opens the module menu for the same module. My preference would be to display the title of the selected menu item instead of the module title in this case.

PR #1408, opened 2019-08-21
Bar code scanning polymer 3

01es:
Bar code scanning support for Android.


01es:
This functionality was delivered in a different way.

PR #1434, opened 2019-09-19
Issue #1427

### LINKED ISSUE #1427: Entity Centre: support OR grouping
### Description

At the moment, conditions for properties that are exposed as selection criteria are a combined with logical AND. The ability to group conditions with logical OR would significantly empower end users in their ability to interrogate the data.

The concept of selection criteria "groups" should be developed to enable users to combine conditions for selection criteria with logical OR by specifying the group number as a mnemonic on a selection criterion. Selection criteria that have the same group number specified belong to the same group and the filtering conditions defined for these selection criteria are to be combined with OR. Naturally, each group would need to have their "brackets" and combined with the rest of conditions using AND. Effectively, it can be thought as if each condition belongs to a different group, unless specified otherwise by the user.

It is difficult to imagine that users would need to use a large number of groups. Therefore, it is proposed to limit the number of group to 9 with the group number ranging from 1 to 9. Such constraint should be used to facilitate entry of group numbers by users as well as to visualise the specified groups.

The sketch below shows how the mnemonic "icon" gets changed when mnemonics are specified. The last "state" is a circle with the group number. This representation should be used when a group number is set, and even if other mnemonics are also specified. Selection criteria with the same group number would be easily identified.

image

At this stage, a simple integer editor can be used for entering group numbers, but with appropriate range validation.

The group mnemonic is not applicable for `@CritOnly` properties as they already have their specific purpose in the underlying query model.

### Expected outcome
The ability by end user to combine search conditions with logical OR.

jhou-pro:
resolve #1427


01es:
As discussed in the Slack channel, the need for further improvements was identified. Pending further discussion.

image

PR #1448, opened 2019-10-30
Issue #938

### LINKED ISSUE #938: Entity Centre: support in-place editing (part 1)
### Description

The discussion below coverer the complete functionality for in-place EGI editing. However, from the implementation perspective, this issue should only be concerned with the single mode editing where a single entity instance represented by a single row gets edited, validated and saved. Once this issue is implemented, issue #1442 should be used to implement the batch mode.

#### Background info
The very original discussion and ideas pertaining to in-place editing of entities in EGI goes back to mid 2015 and can be found [here](https://github.com/fieldenms/tg/wiki/In-Place-Editing-and-Embedded-Entity-Masters). In-place (aka inline) editing of grid controls on the Web is not as ubiquitous as for desktops, and we have not observed any particularly good UX approaches to this since our research in 2015. However, there are real use cases where this kind of data entry is preferred for Desktop users of Web applications.

More recent examples of some usable approaches include components from [Telerik](http://demos.telerik.com/kendo-ui/grid/editing) and [Vaadin](https://demo.vaadin.com/sampler/#ui/grids-and-trees/grid/item-editor). 

Both these approaches rely on "tall" grid rows to hide the transition from between read and edit modes. Unfortunately, tall rows are not suitable for Enterprise grade applications that have to display a lot more information than can fit comfortably with rows of that height. 

The Vaadin's approach uses an overlay that incorporates the editing controls. This approach impedes the usability by disabling both horizontal and vertical grid scrolling. Also, it seems to be applicable strictly to edit/save of values on a single row.

The Telerik's approach is better -- it does not rely on an overlay, navigation between rows and columns is simple and convenient. The modified cells are indicated, and batch saving is supported. However, similarly as with Vaadin's component, such approach is possible only due to a relatively large row height in the read mode matching to row height in the edit mode (i.e. not many rows can be visible at the same time).

#### Design approach
It seems that our original thinking along the line of "expanding" the row that needs to be edited should yield the best result. The grid can remain densely populated ("thin" rows). And only if the row needs to be edited would it transform inline, becoming taller to accommodate taller edit controls and pushing the rows underneath a bit further down. Both horizontal and vertical scrolling would remain fully operational. The edit controls and error reporting would remain exactly the same as for Entity Masters. The validation of edited entities can be individual and asynchronous, while saving can happen in a batch as a single transaction.

The image below is a sketch that illustrates a scenario where an entity in row 4 is being edited. Entity in row 2 has been edited -- two modified properties are indicated with a vertical ember stripe on the right side of each cell. Entity in row 7 -- a single modified property is indicated with vertical ember stripe. Buttons SAVE is enabled indicating that there are unsaved changes.

egi-editing

Saving should proceed entity by entity, but in a single transaction. Saving changes may result in a validation or some other error. Therefore, the firs entity with validation error would have its indicators (stripes on the right side of a cell) turned into red. It could even be a good idea to automatically turn a corresponding row into an edit mode and focus the first property in error. It should be possible to cancel changes for individual entities (a small cancel action may nicely fit on the left side of the edit panel, replacing the edit action). 

The CANCEL button should lead to cancelling of all changes on the current page. Filtering, refreshing or navigating between pages or away to some other menu item in the same compound entity master should be restricted if the current page contains unsaved changes.

Editing cues and keyboard integration for navigation, cancelling and accepting changed is a key to make this design successful. 

1. Hovering over data cells in a row in the editable EGI should "outline" the whole row as if suggesting to click to start editing. Such "outlining" should be done in accordance with the Material Design where a row becomes like a separate stipe of paper placed on top of EGI "sheet" (proper shadows etc.).

2. Double clicking (clicking is reserved for cell actions) any data cell should transform the row into the edit mode, and the first editable property focused.

3. In the edit mode, `Tab` / `Shift + Tab` to navigated between entity editors in the currently edited row. 

4. Tabbing off the last editable property in the row should accept changes, transform the current row into the view mode, make the next row editable (if there is one) and focus to the first editable property there.

5. Shift-tabbing off the first editable property in the row should accept changes, transform the current row into the view mode, make the previous row editable (if there is one) and focus to the last editable property there.

6. In the edit mode, pressing `Enter` in the currently focused property should accept changes, transform the current row into the view mode, make the next row editable (if there is one) and focus the editor, which corresponds to the same property where `Enter` was pressed (this behaviour is needed to support editing of values in a single column).

7. In the edit mode, pressing `Esc` should discard any changes that were made to any property since the last time the row was transformed into the edit mode (any previously accepted changes should remain) and the row should be transformed into the view mode.

### Expected outcome

Reliable and convenient editing of entities that are represented in EGI.

oleh-maikovych:
resolve #938


01es:
@jhou-pro I'll be working on this one today too. Please let me know if there are any changes that you'd like to push.

01es:
A couple of additional items to improve:

- [ ] Swap save and cancel to be in sync with the standard layout for masters.

- [ ] Adjust the tooltips to read "Save changes" and "Cancel changes".

image

PR #1488, opened 2020-02-14
#1482 Converted Authenticator to use user session versioning instead …

### LINKED ISSUE #1482: User session: improved versioning for authentication tokens
### Description

The authentication mechanism in TG is based on the idea of mutating authenticator tokens that get regenerated upon each request. This significantly reduces longevity of authenticator tokens, which in turn reduces a likelihood for these tokens to be used for adverse purposes.

However, web browsers maintain multiple cookies with the same name, which get sent as part of requests until they get expired. In order to identify a cookie that represent the latest authenticator token, it is necessary to version them and select the one with the latest version at the server side.

The current versioning approach is based on the expiration time for authenticator tokens where a token with the latest expiration time is considered the newest. This approach, however, has a problem which manifests intermittently and mainly in conjunction with SSE, where such "newer" tokens in fact represent already stale ones. Possible causes of this might include the use of non-monolithic time-of-day clock, which can "jump" back in time, and async nature of requests from the same users.

Versioning of authenticator tokens should be based on a monolithic (ever growing) mechanism to avoid such situations. The database versioning, which is already in place, should be used instead of expiration time.

### Expected outcome
More reliable monolithic versioning of mutating authentication tokens.

01es:
Resolve #1482

### DIFF
--- a/platform-dao/src/main/java/ua/com/fielden/platform/security/session/UserSessionDao.java
+++ b/platform-dao/src/main/java/ua/com/fielden/platform/security/session/UserSessionDao.java
@@ -43,6 +43,7 @@ import ua.com.fielden.platform.utils.IUniversalConstants;
 @EntityType(UserSession.class)
 public class UserSessionDao extends CommonEntityDao<UserSession> implements IUserSession {
     private final Logger logger = Logger.getLogger(UserSessionDao.class);
+    private static final DateTimeFormatter formatter = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss.SSS");
 
     /** A key to be used for hashing authenticators and series ID before storing them. */
     private final String hashingKey;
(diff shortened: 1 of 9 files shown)


platform-dao/src/main/java/ua/com/fielden/platform/security/session/UserSessionDao.java
```diff
@@ -295,12 +291,13 @@ private String seriesHash(final String seriesId) {
             // remove all user sessions if theft scenario should be considered
             if (shouldConsiderTheftScenario) {
                 // if the session was not found in the cache then proceed with the theft story...
-                logger.warn(format("A seemingly correct authenticator %s did not have a corresponding sesssion record.", auth));
+                logger.warn(format("A seemingly correct authenticator [%s] did not have a corresponding sesssion record.", auth));
```
elvin-fms: Realllly?

platform-dao/src/main/java/ua/com/fielden/platform/security/session/UserSessionDao.java
```diff
@@ -295,12 +291,13 @@ private String seriesHash(final String seriesId) {
...
                 // in this case, sessions are removed based on user name and series ID, which is required taking into consideration that series ID could have been already regenerated
                 final int count = clearAllFoUserAndBySeriesId(user, auth);
                 logger.debug(format("Removed %s session(s) for series ID %s", count, auth.seriesId));
                 return Optional.empty();
             } else { // otherwise just return an empty result, indicating no user session could be found
+                logger.warn(format("A seemingly correct authenticator [%s] did not have a corresponding sesssion record, access denied (skip regeneration == %s).", auth, skipRegeneration));
```
elvin-fms: Realllly?

PR #1560, opened 2020-06-27
Issue #1517

### LINKED ISSUE #1517: BCE/ACE handlers to support runtime parametrisation via entity properties
### Description

Before and After change event (BCE/ACE) handlers, which are mainly used for property validation and post-property-change as part of the model integrity constraints, already support flexible parameterisation. This makes it possible to create quite generic BCE/ACE handlers (e.g. `GreaterValidator` and `MaxValueValidator`).

However, current parameterisation does not support context-dependent parameterisation where parameter values could be specified at runtime depending on some specific logic. Our analysis suggest that having the ability to specify parameters as property values of the same entity as the properties being validated, would provide the flexibility needed. Effectively, such BCE/ACE handler parameters would accept the name of a property that should be read during validation at runtime.

As a working example, consider entity `CentreConfigUpdater`, which represents an EGI configuration for Entity Centres. It has property `pageCapacity` that users can customise to have more or less entries on a single data page. But, we need to ensure that `pageCapacity` would not exceed some upper limit.
For different Entity Centres this upper limit could be different. So, we could not simply declare 

```@Handler(value = MaxValueValidator.class, str = {@StrParam(name = "limit", value = "300")}```,

where `300` is an arbitrary upper limit. 

Instead we would like to have something like 

```@Handler(value = MaxValueValidator.class, prop = {@PropParam(name = "limitPropName", value = "maxPageCapacity")}```, 

where `maxPageCapacity` would be the name of another property in `CentreConfigUpdater`, assigned at runtime in a context-dependent way (e.g. from an Entity Centre configuration).

This issue covers the development of  `@PropParam` and its support by `@Handler`.

### Expected outcome
Runtime parameterisation of BCE/ACE handlers via entity properties.

### PR DESCRIPTION
**Pull request by 01es**
Resolve #1517

### CODE REVIEW

File: platform-pojo-bl/src/main/java/ua/com/fielden/platform/entity/annotation/mutator/PropParam.java [hunk 1]
```diff
@@ -0,0 +1,30 @@
+package ua.com.fielden.platform.entity.annotation.mutator;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Represents a parameter for BCE and ACE handlers that references a property name of the same entity as the property with the corresponding BCE or ACE handler.
```
elvin-fms, line 9: #1517 Only mentions BCE. Should the issue be updated to cover the wider scope of the actual implementation?

01es: @elvin-fms PropParam was intended for  both ACE and BCE parameterisation, but the implementation covered only BCE handles (as per issue description). However, not too much effort was required to extend this support ACE handlers. Such support has now been implemented and issue description enhanced to reflect it.

PR #1561, opened 2020-06-27
Issue #900

### LINKED ISSUE #900: Entity Centre: support custom number of visible and retrievable entities
### Description

It is already possible to set the number of entities to be retrieved and the number of entities to be displayed on an EGI page at the Entity Centre API level (`.setPageCapacity(int)` and `.setVisibleRowsCount(int)`. This feature is very useful and should be made accessible to users.

The EGI customisation feature needs to be enhanced to include the number of entities to be retrieved and displayed. These custom values should be validated and persisted with the rest of custom Entity Centre configuration options. Validation should include:

- [x] 1. The number of visible entities should not exceed the number of retrieved entities.
- [x] 2. The number of visible entities should not be negative. Value `0` should indicate "display all on page".
- [x] 3. The number of retrieved entities should not be less than `1`.
- [x] 4. The number of retrieved entities should not be greater than the maximum number allowed by the default Entity Centre configuration. This requires an enhancement to Entity Centre API to provide API `.setMaxPageCapacity(int)`. The max page capacity should be defaulted to some reasonable value (e.g. `300`) if not specified explicitly.

### Expected outcome
There are many situations where adjusting the number of retrieved and visible in EGI entities would significantly enhance user experience. There are displays with different resolutions where more or less than the default number of EGI rows could be displayed. There are analytical Entity Centres that display charts/graphs representing the data on the current EGI page, and users may wish to increase the amount of information displayed in the chart, but not in EGI (e.g. planning and scheduling feature in TG Air).

Having the number of retrieved and visible EGI rows customisable by users will enable the above and other, currently unforeseen, use cases.

01es:
Resolve #900 

This PR must be merged after PR https://github.com/fieldenms/tg/pull/1560.


platform-pojo-bl/src/main/java/ua/com/fielden/platform/web/centre/CentreConfigUpdater.java
```diff
@@ -33,12 +35,12 @@
...
     @IsProperty(CustomisableColumn.class)
     @Title("Customisable Columns")
-    private Set<CustomisableColumn> customisableColumns = new LinkedHashSet<>();
+    private final Set<CustomisableColumn> customisableColumns = new LinkedHashSet<>();
 
     @IsProperty(value = String.class)
     @Title(value = "Sorting values", desc = "Values of sorting properties -- 'asc', 'desc' or 'none' (the order is important and should be strictly the same as in 'sortingIds' property)")
     @AfterChange(CentreConfigUpdaterSortingValsDefiner.class)
-    private List<String> sortingVals = new ArrayList<>(); // this list should not contain duplicates, please ensure that when setSortingVals invocation is performing
+    private final List<String> sortingVals = new ArrayList<>(); // this list should not contain duplicates, please ensure that when setSortingVals invocation is performing
```
elvin-fms, line 43: If this list should not contain duplicates, it is possible to replace it with `Set`, especially considering that in the only place where `setSortingVals()` appears to be invoked, it collects sorting vals using `Collectors.toCollection(LinkedHashSet::new)`?

01es: @elvin-fms it's a good question. `Set` does not contains duplicates by definition, but by the same definition it does not guarantee any order of its elements. An order in which elements were added is important in this case. The only set-like structure in the standard Java library suitable for this is `LinkedHashSet`. However, it is a bit challenging to use it as the type for collectional property (e.g. `Collections.unmodifiableSet` cannot return an instance of `LinkedHashSet`). If `Set` is used, then any information about the order of elements is not conveyed at the property type level. But order is more important in this case than the absence of duplicate elements. Hence, most likely, the approach chosen with `List`. Of course, the use of some kind of an immutable set implementation with the right order would be ideal, but there are none that we currently support adequately (e.g. Guava collections could potentially be used, but these types are not supported as property types).

PR #1566, opened 2020-07-09
Issue #1554

### LINKED ISSUE #1554: Global Error Handler
### Description

Currently errors are handled gracefully in Entity Centres and Masters. Some additional work was done to handle errors in `tg-ui-action` instances that are used on Entity Centre or Master. But not all errors are handled and some errors may be unexpected. This is why a global error handler that could catch untapped exceptions is required.

Such global handler should be able to restore the state of components if the caught exception provides the relevant information for that (e.g. stop a spinner and enable a button), and display a toast message about the error if applicable.

It is also desired to report client-side errors back to the application server. This way the application would collect errors that users may not see or may decide not to report.

### Expected outcome
Graceful error handling with the ability to recover the state of the client application in order to make it usable. And the ability to collect client-side errors, enabling faster and more reliable application improvement cycle.

oleh-maikovych:
resolve #1554


platform-web-ui/src/main/web/ua/com/fielden/platform/web/sse/tg-sse-behavior.js
```diff
@@ -139,9 +139,7 @@ export const TgSseBehavior = {
                 }, 15000);
             }
 
-            if (!self.errorHandler) {
-                throw new UnreportableError(e);
-            }
+            throw new UnreportableError(e);
```
01es, line 142: Why this check was removed? What if the custom error handler already throws new `UnreportableError(e)`?

platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-global-error-handler.js
```diff
@@ -0,0 +1,127 @@
...
+
+export class UnreportableError extends Error {
+    constructor(...params) {
+        super(params);
+
+        if (Error.captureStackTrace) {
+        Error.captureStackTrace(this, UnreportableError);
+        }
+
+        this.name = "UnreportableError";
```
01es, line 16: What is the purpose of `this.name`? If it is really needed, why not change this assignment to `this.name = UnreportableError .name;` ?

platform-web-ui/src/main/web/ua/com/fielden/platform/web/sse/tg-sse-behavior.js
oleh-maikovych: Yes, it is possible that custom `errorHandler` may throw `UnreportableError`, but now only `tg-file-processing-behavior` provides custom `errorHandler` and it doesn't throws any errors. Should this `UnreportableError` be thrown in `tg-file-processing-behavior`

platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-global-error-handler.js
oleh-maikovych: If `this.name` is specified then the error on server will look like this:
```
Uncaught UnreportableError: [object Event] Error happened in: https://tgdev.com:8091/resources/sse/tg-sse-behavior.js at Ln: 142, Co: 13
UnreportableError: [object Event]
    at EventSource.<anonymous> (https://tgdev.com:8091/resources/sse/tg-sse-behavior.js:142:19)
```
otherwise it will look like this:
```
Uncaught Error: [object Event] Error happened in: https://tgdev.com:8091/resources/sse/tg-sse-behavior.js at Ln: 142, Co: 13
Error: [object Event]
    at EventSource.<anonymous> (https://tgdev.com:8091/resources/sse/tg-sse-behavior.js:142:19)
```
Therefore `this.name` provides additional information about the type of `Error`.
Indeed `"UnreportableError"` can be replaced with  `UnreportableError.name` but if somebody changes the name of class then the one will get an `undefined` error

platform-web-ui/src/main/web/ua/com/fielden/platform/web/sse/tg-sse-behavior.js
01es: I think that if a customer error handler is provided then there is no need to throw an error. If a custom error handler throws an exception then it would break the reconnection logic.

So, my question remains -- why this check was removed?

platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-global-error-handler.js
01es: And if someone changes the class name and does not update the assignment to `this.name` they would get an incorrect error name in the log.

Should the web unit tests not catch the "undefined" error?

01es: @oleh-maikovych perhaps it would be possible to use `error.constructor.name` instead of `error.name` and remove `this.name`?

oleh-maikovych: Yes this can be done and the [commit](https://github.com/fieldenms/tg/commit/1a5892fb6432e6ecf277d6b66b3e5fdf285b65e0) shows how it was done

platform-web-ui/src/main/web/ua/com/fielden/platform/web/sse/tg-sse-behavior.js
oleh-maikovych: Further discussion took  place in Slack channel and it was decided to get rid of error throwing in `tg-sse-behavior` this [commit](https://github.com/fieldenms/tg/commit/5a50ba039e2d4e08fcbfa187a5cc8d6eba8f7487) hopefully resolves this conversation

PR #1571, opened 2020-07-16
Issue #1570

### LINKED ISSUE #1570: Locator doesn't open chosen entity
### Description

Entity Locator allows one to open entity master for selected entity. Because of recent changes (#1533) to  DOM dialog management logic, this functionality stopped work properly. Sometimes it throws error, saying that `retrieve` is not a function, or doesn't open dialog at all. Preliminary investigation suggests that locator's `_dialogClosed` method was invoked after entity master's `showDialog` . This caused problem in logic that adds and removes dialog to document DOM.

### Expected outcome

Locator should allow one to select desired entity and open entity master for it.

### Actual outcome 

Locator sometimes fails to open entity master for desired entity.

oleh-maikovych:
resolve #1570

### DIFF
--- a/platform-web-ui/src/main/web/ua/com/fielden/platform/web/centre/tg-entity-centre-behavior.js
+++ b/platform-web-ui/src/main/web/ua/com/fielden/platform/web/centre/tg-entity-centre-behavior.js
@@ -819,7 +819,7 @@ const TgEntityCentreBehaviorImpl = {
         /* Provide predicate for egi that determines whether inline master can be opened or not.
          * It can not be opened if another master in dialog is opened. */
         this.$.egi.canOpenMaster = function () {
-            return !this.actionDialog.opened;
+            return this.actionDialog === null || !this.actionDialog.opened;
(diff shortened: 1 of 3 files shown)


01es:
Interactive testing was successful. However, I'd like to better understand why there are "many" places where the removal of the dialog from DOM needs to be invoked.

oleh-maikovych:
> Interactive testing was successful. However, I'd like to better understand why there are "many" places where the removal of the dialog from DOM needs to be invoked.

oleh-maikovych:
Removing dialog from DOM happens in three places.
1. Normal dialog closing,  when user saves entity master and dialog closes or user presses 'x' button on right top corner.
2. No-UI masters also invoke `showDialog` method which causes dialog to be added to DOM as the result this dialog should also be removed from DOM after No-UI action master execution.
3. During dialog opening some error happened which caused dialog to stay closed, but it was added to DOM in `showDialog` method, in that case this dialog should also be removed from DOM.

PR #1586, opened 2020-08-17
#1585 - implemented sanitizing of String parameters for Jasper reports

### LINKED ISSUE #1585: Sanitizing of String parameters in all jasper reports
### Description
It was reported that Jasper reports are producing sub-optimal output due to the presence of tabs in some of the `String` values.

"Helper" class `EntityAggregatesReportSource` is used as an intermediate data container for Jasper reports.
By "sanitizing" `String` values that go through this class, it would be possible to improve the output of a large proportion of Jasper reports - i.e. those that that use `EntityAggregatesReportSource`.

Such sanitizing should:

- [x] 1. Replace tabs and various non-breaking spaces with regular spaces.
- [x] 2. Trim any non-printable characters: spaces (including those that appeared after replacing in step 1), line breaks and carriage returns, from both ends of `String` values.

### Expected outcome
Reduced effect of non-printable characters on the output of printable reports.

julia-fms:
resolve #1585

### DIFF
--- a/platform-pojo-bl/src/main/java/ua/com/fielden/platform/file_reports/ReportUtils.java
+++ b/platform-pojo-bl/src/main/java/ua/com/fielden/platform/file_reports/ReportUtils.java
@@ -1,0 +1,27 @@
+package ua.com.fielden.platform.file_reports;
+
+import java.util.Optional;
+
+import org.apache.commons.lang.StringUtils;
+
+public final class ReportUtils {
+
+    // Prevent creating instances of this class
+    private ReportUtils() {
+
+    }
+    
+    /**
+     * Sanitize String parameters for Jasper reports to replace horizontal whitespace character (including tabs) with regular spaces.
+     * Also trim non-printable characters, including tab and non-breaking spaces as well as line breaks and carriage returns from both the start and the end.         
+     *
+     * @param dirty
+     * @return clean
+     */
+    public static String sanitizeString(final String dirty) {
+        final String dirtyNotNull = Optional.ofNullable(dirty).orElse("");
+        // "\\h" handles tabs as well as all sorts of weird non-breaking spaces, but not line breaks.
+        return StringUtils.trimToEmpty(dirtyNotNull.replaceAll("[\\h]+", " "));
+    }
+
+}
--- a/platform-pojo-bl/src/main/java/ua/com/fielden/platform/file_reports/EntityAggregatesReportSource.java
+++ b/platform-pojo-bl/src/main/java/ua/com/fielden/platform/file_reports/EntityAggregatesReportSource.java
@@ -40,6 +40,7 @@ public class EntityAggregatesReportSource implements JRRewindableDataSource {
             throw new JRException("Invalid field: " + jrField.getName());
         }
 
-        return EntityUtils.toString(data[index].get(jrField.getName()), props.get(jrField.getName()));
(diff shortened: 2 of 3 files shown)


platform-pojo-bl/src/main/java/ua/com/fielden/platform/file_reports/ReportUtils.java
```diff
@@ -0,0 +1,27 @@
...
+    }
+    
+    /**
+     * Sanitize String parameters for Jasper reports to replace horizontal whitespace character (including tabs) with regular spaces.
+     * Also trim non-printable characters, including tab and non-breaking spaces as well as line breaks and carriage returns from both the start and the end.         
+     *
+     * @param dirty
+     * @return clean
+     */
+    public static String sanitizeString(final String dirty) {
```
01es, line 21: This method converts `null` values into empty strings. This may lead to undesired side effects. Returning `Optional<String>`, with empty value for `dirty == null`, would let the user of this function to decide how "empty" values should be treated.

julia-fms: Jasper report provides way to handle `null`, but it is not done by default.

I will try to demonstrate this here and take Job Pack for Ixom as example.

image

If value is `null` it will simply print word **null** as output. 

To avoid it "Blank when null" should be checked:

image

and the result will look like this (blank for Originator):

image

And in most cases it is not checked.

To be on the safe side I often pass blank string to avoid printing of "null" word by simply checking for nulls: 
```Java
dataEntry.put("supplierPostalAddressLine1", supplier.getStreetAddress() == null ? "" : supplier.getStreetAddress().getStreet());
```
This why I decided to handle `null` values as empty strings as part of `sanitizeString` method.

01es: The intent was understood. The change request will give more control to the end developer and to avoid any surprises. For example, someone may want to explicitly state "Missing value" for `nulls`. For the demonstrated case, you would simply say `sanitizeString(...).orElse("")`.

01es:
@nazchk after additional analysis and realisation that this is a platform-level issue (I mistakenly though that it is an application -level issue due to its original description), I think that these changes do not really belong to the platform. My preference would be for such value transformations to be implemented at the application level. For example, there could be class `RemoveNonPrintableCharsReportSource` at the application level, extending `EntityAggregatesReportSource`, that would do the necessary value transformations, and instances of such class would then be used in all places where `EntityAggregatesReportSource` is used currently.

I have discussed this with @elvin-fms and @julia-fms. However, @elvin-fms does not agree. Therefore, please consider and share your opinion regarding the implemented and suggested in post approaches.

elvin-fms:
@01es 

Where did @elvin-fms did not agree?

To your suggestion of `NoNonPrintableCharsReportSource` via Slack, @elvin-fms has replied the following:

> Так, як зараз є у цих двох ПР, виглядає, що вдалося досягти максимального ефекту мінімальними змінами.
> Мабуть можна досягнути подібного ефекту і більшими змінами.
> Якщо виносити код повязаний з Джаспером з платформи у якусь бібліотеку, можливо варто також глянути і до AbstractPdfReportFactory - таке враження, що він у TG Air не використовується, а може було б варто.

In my understanding, **можна** implies agreement.

elvin-fms:
As discussed and agreed with @01es, processing of `String` values passed to JasperReports will be performed on application level.

PR #1616, opened 2020-11-10
Issue #1607

### LINKED ISSUE #1607: Multi-action: support for context-dependent actions
### Description

In practice there are situations where the same action should have very different behaviour depending on the state of an entity it is invoked on. For the purpose of this discussion we can constrain ourselves to Primary and Secondary Actions on Entity Centre. However, the same would also be applicable to Property Actions on Entity Masters.

Let's use a practical scenario of Inventory Returns for Work Orders that is represented as a secondary action on the embedded centre for Work Order Master. There are two possible situations -- if there is only one inventory item issued then Inventory Return should be handled by action `SingleItemInventoryReturn`, and if there are multiple items issued then such returns should be  handled by action `MultiItemInventoryReturn`. The choice between the two actions is context dependent where the context is an entity instance for which an action gets invoked, and in this case, is driven by the number of inventory items issued (i.e. some property value for the current entity).

Currently such functionality can be achieved, but it is cumbersome and error prone. A special support for what we now refer to as "multi-action" would establish a standard way of implementing this functionality in a simple and elegant way, while preserving low coupling and reusing already established concepts of functional entities, actions and contexts.

There are three main elements that need to be developed to support multi-actions:

- [x] 1. Extended Action DSL that would support configuration of multi-actions `EntityMultiActionConfig` defined by an action-selector and a list of actions (instances of `EntityActionConfig`). Action-selector would be specified as a class name that implements the required contract, similar in nature to the "rendering customiser", but specified at the level of `EntityMultiActionConfig`. The main role for action-selectors is to determine what action out of the specified ones should be presented depending on the current entity.

  Entity Centre DSL would need to support for `addPrimaryAction` or `addPrimaryMultiAction` (either or), `addSecondaryAction` and `addSecondaryMultiAction` (due to the nature of secondary actions where there could already be multiple different actions, each of the actions in the list could be either be an action or a multi-action). The new DSL should be introduced in a fully backward compatible manner.

- [x] 2. Processing of action-selectors, which should be implemented in a way similar to how rendering customisers get processed. It is necessary to establish a metadata about the "current" action that would be marshalled to the client together with the actual data to be rendered by EGI.

- [x] 3. Client-side processing of the multi-action metadata to correctly render and invoke actions from the list of a multi-action on per-instance basis.

### Expected outcome
Simple and powerful abstraction for context dependent multi-actions on Entity Centres for Primary and Secondary roles.

oleh-maikovych:
resolve #1607


platform-web-ui/src/main/java/ua/com/fielden/platform/web/centre/EntityCentre.java
```diff
@@ -816,6 +819,19 @@ public boolean shouldEnforcePostSaveRefresh() {
         }
     }
 
+    public Optional<IEntityMultiActionSelector> getPrimaryActionSelector() {
+        return this.dslDefaultConfig.getResultSetPrimaryEntityAction()
+                .map(multiActionConfig -> !multiActionConfig.isNoAction() ? injector.getInstance(multiActionConfig.actionSelectorClass()) : null);
+    }
+
+    @SuppressWarnings("unchecked")
+    public Optional<List<IEntityMultiActionSelector>> getSecondaryActionSelectors() {
```
01es, line 828: No need to wrap containers into Optional. An empty container should be returned in case of no data.

jhou-pro (approved):

jhou-pro:
Superseded by #1638.

PR #1623, opened 2020-12-15
Issue #1534

### LINKED ISSUE #1534: Entity Centre: support to share configurations between users
### Description

Configurations for Entity Centres are uniquely identifiable. Access to Entity Centres is controlled via on per user role basis. This makes it possible for users to share their saved configurations via URI by specifying `?UUID=value`, which can be opened with HTTP GET.

Action image would need to be made available as part of the Entity Centre selection criteria panel. This action should generate a URI and simultaneous copy that URI into [clipboard](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard). Generation of URI should only be possible for saved-as and saved configurations.

The generated in this way URI could then be used to share with other users that can access to the same Entity Centre. Users who loaded an Entity Centre via such URI should be able to save that configuration as their own using action "SAVE".

### Expected outcome
Ability to securely share Entity Centre configurations via URI with other users.

jhou-pro:
resolve #1534


oleh-maikovych (approved):

jhou-pro:
Superseded by #1638.

Issue #1661, opened 2021-03-30
Bijection between client- and sever-side date/time formats


01es:
#### Background
In most cases formatting of dates at the client- and server-sides is independent and may follow different formats. This is due to a fact that date/time value between the client- and server-sides are marshalled as the number of milliseconds.

However, there are cases where date/time value are transmitted as formatted strings. This happens, for example, when a property of type `Date` represents one of the composite key members for an entity that needs to be autocompleted. Autocompleters operate strictly with string representations and try to match and reconstruct entities by the string representation of their keys.

String representation of date/time values at the client side are different than those used at the server-side. More to the point, the date and time formatting patters for converting date values to the same string representations, are different between JS, Java and various supported RDBMSes.

#### Requirements
This issue covers the work required to rectify the differences between different time formats at different application layers -- JS at the client, Java at the application server, RDBMSes at the DB level.

The following items outline the scope of work:

- [ ] Map date and time formats (separately) between JS, Java and SQL for supported RDBMSes (H2, SQL Server and PostgreSQL).

- [ ] Need to introduce a predefined set of date and time formats to be supported. This is only reasonable taking into account that each format needs special support at the client-side for approximations as well as mapping between different app layers.

- [ ] `IDates` should drive formatting, format mapping and setting of an active format for an application. As a note, this issue is not concerned with time-zone specific transformations, a separate issue would need to be created for that.

- [ ] `IDates` need to be incorporated into `EntityUtils` and `DynamicEntityKey` to be used in respective methods `toString` when converting `Date` values to string.
    _Implementation remark:_ A consideration should be given to using Guice's ability to inject static variables. If this approached is considered suitable then `requestStaticInjection(EntityUtils.class);` and `requestStaticInjection(DynamicEntityKey.class);` would need to be added to `BasicWebServerModule`, with fields `@Inject private static IDates dates;` added to both `EntityUtils` and `DynamicEntityKey`.

- [ ] EQL's `AbstractFunction` is responsible for converting instances of `ISingleOperand` to string and provide special handling for dates. This class needs to be enhanced to accept `IDates` and implement conversion of dates in accordance with mappings between date and time formats. SQL Server conversion should used function `FORMAT` instead of `CONVERT`. Class `AbstractFunction` is used in `Concat.sqlForH2AndMssql` and `Concat. sqlForPostgresql`.

[...]

PR #1711, opened 2021-06-26
Issue #1676

### LINKED ISSUE #1676: Web Resources: improved time zone awareness
### Description

Currently Web Resources only assign the time zone information for a request if that information for provided. In the context of a Web UI client this is tightly controlled and a time zone is always provided. Now that other clients are supported by means of a generic GraphQL and specialised REST integration, it became important to provide a more robust fall-back approach in cases where a client time zone was not provided.

A request time zone information is stored as a `ThreadLocal` variable, which is assigned each time a request comes in. However, if a request carries no time zone information, by mistake or deliberately, then no value is assigned. As a result, if a thread that had time zone assigned previously is reused for request handling (due to thread pooling), then a potentially invalid time zone for that request is also reused. This is an uncertain situation bordering on a defect and needs to be addressed.

There are 2 cases:

1. There are web resources that do not set a request time zone deliberately, because they should/intend to work in the default server time-zone.

2. There could be clients other than a TG-bed Web UI that assume the use of a default server time zone, and deliberately do not pass any time zone information.

A "soft" approach to addressing this situation would be to assign the default server time zone as a request time zone, if the request does not carry a time zone. A "hard" approach would be to always return an error for a request carrying no time zone information. This might be a configuration option. These two approaches need to be discussed further.

### Expected outcome
Deterministic handling of requests without time zone information.

jhou-pro:
resolve #1676


01es (commented):
@jhou-pro could you please ascertain whether `requestStaticInjection` for `TgScalars` could be used – this should significantly reduce the ritual of passing `IDates` instances throughout the GraphQL related code. Such binding would need to be different for application and testing runtimes.
We have discussed the approach with using `requestStaticInjection` for `IDates` as part of https://github.com/fieldenms/tg/issues/1661

01es (approved):

PR #1719, opened 2021-07-21
Issue #1718

### LINKED ISSUE #1718: Further enhancements to Eclipse TG templates
### Description
TG Platform provides convenient and useful code templates, which facilitate adding new properties to entities and the creation of EQL queries.

As the platform evolves, these templates also need to evolve to stay relevant and up-to-date:

- [x] 1. Create templates `tgprop-string` and `tgprop-composite-key-member-string`, which will include `(length = 255)` and `@BeforeChange(@Handler(MaxLengthValidator.class))`.
- [x] 2. Modify `tgprop-collectional` to replace `HashSet` with `LinkedHashSet`.
- [x] 3. Add annotation `@Title` as commented - in most cases the titles can be derived from property name or type.

### Expected outcome
Relevant and up-to-date TG templates.

elvin-fms:
Resolve #1718

### DIFF
--- a/platform-doc/eclipse-templates/tg-templates.xml
+++ b/platform-doc/eclipse-templates/tg-templates.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?><templates><template autoinsert="true" context="java" deleted="false" description="Generates a stub for a calculated property" enabled="true" name="tgprop-calculated">@IsProperty
 @Readonly
 @Calculated
-@Title(value = "${Title}", desc = "${Desc}")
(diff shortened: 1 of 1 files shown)


01es (changes requested):
I would not recommend commenting out @Title as engineers should be encourage to provide extended description for each property. And if @Title is not required then it is a trivial matter to delete it -- as trivial as to delete a commented out @Title, if the title is not required. However, if @Title is commented out in such a case, there is less incentive to delete it and a commented out code could remain by mistake.

elvin-fms:
Encouraged

01es (approved):

PR #1762, opened 2021-11-04
Issue #1761 (breaking change, apps need to be updated)

### LINKED ISSUE #1761: Boolean properties: requiredness integrity constraint
### Description

Boolean properties have an ordinary type `boolean`, which means they always have a value – `false` or `true`. Making such properties required either at the declaration with `@Required` or dynamically at runtime does not really have any semantics as it would always have some non-null value.

Based on several practical use cases and a number of experiments, it now seem reasonable to introduce the requireness semantics for boolean properties. More specifically, required boolean properties much require to have value `true`.

The necessary changes should include:

- [x] Changes to `MetaPropertyFull`, method `validate` and `isValidWithRequiredCheck` to include the check for boolean properties.
- [x] Improve rendering for `tg-boolean-editor` in case of requiredness and validation error by making the checkbox blue or red to match the colour of the underline

#### Remark about composite key members
There are cases where boolean properties need to be used as one of the composite key members for an entity. The default semantics of `@CompositeKeyMember` makes each member required. In application to boolean properties this would mean they must be `true`. In most cases composite key members should be able to accept either `true` or `false`. For this, boolean composite key members should be have optional (relaxed) requiredness, which is achieved by annotating them with `@Optional`.

### Expected outcome
Requiredness semantics for boolean properties, enforcing the use of `true` for required properties.

01es:
Resolve #1761

This change has already been merged into the Modern Java branch.


jhou-pro (approved):

01es:
Already merged into the Modern Java branch.

PR #1776, opened 2021-12-12
#1774 Implemented the core issue requiremetns + tests.

### LINKED ISSUE #1774: Key Number: support generation of a batch of sequential key values
### Description

Entity `KeyNumber` represents a concept of sequential numbers used for generating key values for entities that require such capability. For example, work orders, purchase orders, incidents, all have their business key values generated rather than entered by a human.

At the moment the key numbers are generated one at the time and are underpinned by a database table with a pessimistic lock on the record being used to produce the next number and to update that value in the database.

In practice there are situations where it is necessary to create a batch of new instances with sequential keys. This requires for the key number generation logic to update the next value in a corresponding table not by 1, but by the number of key values required.

Contract `IKeyNumber` needs to be enriched with method `SortedSet<Integer> nextNumbers(final String key, final int count)` to support generation of a batch of sequential key values.

### Expected outcome
Ability to generate a batch of sequential key values.

01es:
Resolve #1774

### DIFF
--- a/platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
+++ b/platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
@@ -2,6 +2,14 @@ package ua.com.fielden.platform.keygen;
 
 import static java.lang.String.format;
 
+import java.util.Collections;
+import java.util.SortedSet;
+import java.util.TreeSet;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+import org.hibernate.LockMode;
 import org.hibernate.LockOptions;
 
 import com.google.inject.Inject;
@@ -37,8 +45,8 @@ public class KeyNumberDao extends CommonEntityDao<KeyNumber> implements IKeyNumb
     public Integer nextNumber(final String key) {
         KeyNumber number = findByKey(key); // find an instance
         if (number != null) {
-            // re-fetch instance with pessimistic UPGRADE lock
-            number = (KeyNumber) getSession().load(KeyNumber.class, number.getId(), LockOptions.UPGRADE);
+            // re-fetch instance with pessimistic write lock
+            number = (KeyNumber) getSession().load(KeyNumber.class, number.getId(), new LockOptions(LockMode.PESSIMISTIC_WRITE));
         } else { // this would most likely never happen since the target legacy db should already have some values in table NUMBERS
             number = factory.newByKey(KeyNumber.class, key).setValue("0");
         }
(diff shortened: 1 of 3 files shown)


platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
```diff
@@ -37,8 +45,8 @@ protected KeyNumberDao(final IFilter filter, final EntityFactory factory) {
     public Integer nextNumber(final String key) {
         KeyNumber number = findByKey(key); // find an instance
         if (number != null) {
-            // re-fetch instance with pessimistic UPGRADE lock
-            number = (KeyNumber) getSession().load(KeyNumber.class, number.getId(), LockOptions.UPGRADE);
+            // re-fetch instance with pessimistic write lock
+            number = (KeyNumber) getSession().load(KeyNumber.class, number.getId(), new LockOptions(LockMode.PESSIMISTIC_WRITE));
```
elvin-fms, line 49: Why create `new LockOptions()` instead of simply passing a single `LockMode`, as before?
Also, the typecast appears unnecessary, Eclipse automatically removes it for me.

platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
```diff
@@ -49,6 +57,27 @@ public Integer nextNumber(final String key) {
...
+    @SessionRequired
+    public SortedSet<Integer> nextNumbers(final String key, final int count) {
+        if (count < 1) {
+            return Collections.emptySortedSet();
+        }
+
+        KeyNumber number = findByKey(key); // find an instance
+        if (number != null) {
+            // re-fetch instance with pessimistic write lock
+            number = (KeyNumber) getSession().load(KeyNumber.class, number.getId(), new LockOptions(LockMode.PESSIMISTIC_WRITE));
```
elvin-fms, line 70: Ditto.

platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
01es: UPGRADE is deprecated.
image

platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
01es: Ditto.

platform-dao/src/main/java/ua/com/fielden/platform/keygen/KeyNumberDao.java
elvin-fms: I have noticed that.
What I mean is – why not write simpler, like this:
```
number = getSession().load(KeyNumber.class, number.getId(), LockMode.PESSIMISTIC_WRITE);
```

01es: This looks like to be an alternative API. The one with with `LockOptions` was used for safety as the previous implementation was working well for years and was tested again after the change to `PESSIMISTIC_WRITE`. The use of an alternative `load` would need to be analysed. One would hope it would work in the same way, but with complex systems like Hibernate, it is best to take extra care. It is propose to leave this as is, as this is just a minor implementation detail.

elvin-fms: Aha! I did not notice that one version uses `LockOptions`, while the other uses `LockMode`.
Sorry for your time.

PR #1798, opened 2022-03-18
Issue #1791

### LINKED ISSUE #1791: Entity Centre: exclude insertion point activation during centre refresh on postActionSuccess event
### Description

After an action on an Entity Centre gets executed and `postActionSuccess` method is invoked, the centre receives a refresh event, which refreshes the centre. This also reactivates every insertion point, associated with this Entity Centre. In most cases it is the intended and expected behaviour, which refresh the data for each insertion point. However, there are situations where it is known that data for some of the insertion points won't be changed, but refreshing is computationally intensive. It is required to provide a way for developers to fine-tune refreshing of insertion points by specifying which ones should not be refreshed.

To implement this issue:

- [x] Add a property to `tg-ui-action` to specify insertion points to be excluded from refresh.
- [x] Enhance the implementation of Entity Centre `currentPage` to accept a list of insertion point names to be excluded, and adjust its implementation to skip reactivation for those insertion points.
- [x] Provide API for entity centre DSL to exclude insertion point from centre refresh cycle of action.

### Expected outcome

Ability to exclude specific insertion points from being refreshed after an execution of action, defined for the main Entity Centre.

oleh-maikovych:
resolve #1791

### DIFF
--- a/platform-web-ui/src/main/web/ua/com/fielden/platform/web/actions/tg-ui-action.js
+++ b/platform-web-ui/src/main/web/ua/com/fielden/platform/web/actions/tg-ui-action.js
@@ -273,6 +273,14 @@ Polymer({
         },
 
         /**
+         * Holds the element-name of entity master for insertion point that should be excluded when refreshing parent entity center after action execution.
+         */
+        excludeInsertionPoints: {
+            type: Array,
+            value: null
+        },
+
+        /**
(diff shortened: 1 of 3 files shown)


01es:
These changes will be merged as part of https://github.com/fieldenms/tg/pull/1799

PR #1799, opened 2022-03-18
Issue #1793

### LINKED ISSUE #1791: Entity Centre: exclude insertion point activation during centre refresh on postActionSuccess event

### LINKED ISSUE #1793: EGI: implement support for a client-side dataset with pagination
### Description

EGI is a data-grid with rich interactive functionality, which supports data pagination.
Data pagination is the main performance-related feature. It ensures that an amount of data loaded from the server is limited by the page size.
The main objectives for this, is to keep the memory utilisation at both the server and client ends low and data transfer over the network fast. Ordinarily, the page size does not exceed 30-50 records, which satisfies these objectives.

However, there are situation in practice where a large amount of data (e.g. 1,000 of records) needs to be send to the client. The most prominent example of this, is data visualisation, such as GIS or charts, which also require for the visualised data to be displayed in EGI. In such cases, EGI has a large data page with all the records rendered at once. This leads to very poor rendering performance due to creation of a large number of DOM elements for displaying the data.

In order to reduce the number of DOM elements for a large dataset, it is to propose to support a client-side data pagination, where a large dataset loaded to the client would not be rendered at once, but split into "pages" as if that data is loaded page-by-page. The mechanism of "next" and "previous" page would navigate over the locally cached data, which was received upon "run" or "refresh". Action "refresh" would treat all the local data as a "single page", and refreshing the data would reload the data as if there is only a single page, but would "jump" to the current local page to be displayed. If that local page would not exist (e.g., fewer records were returned after refresh) then the first page would get displayed.

All other standard EGI behaviour should also treat the whole local dataset as if it was a single "standard" page (e.g., selected items could spread over several "local" pages and exporting selected items would include those). In other words, client-side pagination should be treated strictly as an optimisation technique. Having the page size set to "30" in this case would guarantee a small number of DOM elements for rendering only a small subset of the retrieved data. Pagination would work fast without the need to interact with the server-side, and the data visualisation components (e.g., a map or a chart) would still have access to the whole dataset irrespective of the current page being displayed.

The Entity Centre DSL would need to be enhanced to support new configuration setting `retrieveAll` in addition to `setPageCapacity` and `setMaxPageCapacity`. If `retrieveAll` is specified then `setPageCapacity` and `setMaxPageCapacity` would have a local, client-side meaning. For example, in case of the work order scheduling, `retrieveAll` would ensure that all scheduling data is retrieved without any capacity limit, but displayed in pages limited by the specified page capacity. In this case it would also make sense to restrict the maximum page capacity to the same value as page capacity (e.g., 30 or 50) to ensure fast rendering. The interactive chart, used for scheduling, would display all the retrieved data – not just on the current "page". So, there would need to be a client-side API to access all the data as opposed to only the page data. Having the `retrieveAll` setting would also avoid possible situations where setting up large page capacity could still inadvertently limit the number of the actual data points.

It is estimated that the approach with local "pagination" should achieving the necessary performance characteristics. There would be some minor inconsistency pertaining to actions that operate on "selected" items. For example, the export action would export selected items that were selected on "all" pages as opposed only those that were selected on the "current" page. However, this should "feel" natural for the intended case of "large" datasets.

### Expected outcome
Significantly improved rendering performance for EGI, handling large datasets.

oleh-maikovych:
resolve #1793
resolve #1791


jhou-pro (approved):

01es (approved):
Accepting with a caveat that handling of selected EGI rows for virtual pages needs to be aligned with the current behaviour of non-virtual pages (i.e., when `retrieveAll` is not used). More details are provided in Slack.

PR #1803, opened 2022-04-08
Issue #1790 Improvements to the data migration framework

### LINKED ISSUE #1790: Further improvements to the data migration framework
The following improvements need to be made to `RetrieverDeadReferencesSeeker`:

- [x] correct handling of the case when there is no retriever at all for entity type being used
- [x] correct validation SQL for cases of entity types with optional key members
- [x] correct validation SQL for cases of entity type being retrieved through several retrievers in the scope of their execution

The following improvements need to be made to `RetrieverSqlProducer`:

- [x] correct SQL for uniqueness check (handle `NULL`s properly for optional key members)
- [x] correct SQL for uniqueness check (check union of data of all retrievers for each entity type)

Also include the following:
- [x] replace old `EQL` metadata usage with the new one 
- [x] remove obsolete validations
- [x] implement data integrity validations of updaters keys
- [x] perform the retriever `SQL` syntax validation prior to all other validations

nazchk:
Resolve #1790


platform-dao/src/main/java/ua/com/fielden/platform/migration/DataMigrator.java
```diff
@@ -103,23 +86,38 @@ public DataMigrator(
         LOGGER.info("Migration duration: " + pd.getMinutes() + " m " + pd.getSeconds() + " s " + pd.getMillis() + " ms");
     }
 
+	private static void printRetrieversScheme(final List<CompiledRetriever> retrieversJobs) {
```
elvin-fms, line 89: This method does not appear to be used. Can it be removed?

elvin-fms (changes requested):
@nc

As I was going through this PR, my Eclipse was keen to replace `StringBuffer` with `StringBuilder` in some places (because it is configured so, but only for local variables).

I refrained from accepting this change, partially because I was not 100% sure where data migration is multi-threaded, partially because some instances were not being replaced and partially because the performance gain is probably negligible. Still, I am mentioning this for completeness – please consider using `StringBuilder` where appropriate.

platform-dao/src/main/java/ua/com/fielden/platform/migration/DataMigrator.java
01es: @elvin-fms and @nazchk , `StringBuilder` should always be preferred over `StringBuffer` in the local context (e.g., method calls). More global states of these types, which can be accessed concurrently, should be avoided.

PR #1814, opened 2022-05-22
Issue #1812

### LINKED ISSUE #1812: GIS: layer unselection
### Description
There is a need to be able to unselect layer on a map. In case if it is selected, just perform tap and this should unselect it (including popup). In case if popup was closed before that, then it should not be opened on unselection action.

### Expected outcome
Unselection action to be performed on GIS component.

jhou-pro:
resolve #1812


01es (approved):
A minor improvement is desirable -- blocking the popup from appearing for a brief moment when unselecting a marker. It is not a problem per se, but sometimes a "flickering" effect is noticeable.

PR #1840, opened 2022-06-24
Issue #1818 (OR-grouping for @CritOnly)

### LINKED ISSUE #1818: Entity Centre: OR grouping support for @CritOnly properties
### Description
A bit of background. Issue #947 introduced support for `@CritOnly`-driven conditions at the EQL level (`.critCondition` API). This support was used effectively to enable expression of conditions in application to collectional associations with the use for `@CritOnly(MULTI)` selection criteria (aka collectional criteria), providing support for mnemonics `Missing` and `Not`. Issue #1427 introduced support for OR-grouping for selection criteria in Entity Centres. However, this support was not implemented for collectional criteria.

As the next step in enriching data search capabilities with Entity Centres, it is now required to support OR-grouping for collectional criteria. This is the core requirement for this issue.

##### Single and Range criteria
It has also been observed that in practice, criteria `@CritOnly(SINGLE)` and potentially `@CritOnly(RANGE)` can be very convenient or even required. It is necessary to verify if these criteria kinds can be used with `.critCondition`. Everything discussed here in relation to `@CritOnly(MULTI)` as the base case, should also be considered for `@CritOnly(SINGLE)` and `@CritOnly(RANGE)` -- this should not increase the scope of required work significantly.

#### Implementation details

EQL provides 2 overloaded API `.critCondition` – one without and one with a partial EQL model (`ICompoundCondition0<?>` to be precise). The version that accepts argument a partial EQL model is of our interest. Currently this API is used when defining a model for a Synthetic Entity, and because of this, it necessarily has to be combined with other conditions using either `.or()` or `.and()`, which cannot be changed at runtime. 

As a running example, consider a model below:
```Java
protected static final EntityResultQueryModel<ReWorkActivity> model_ = 
                    select(WorkActivity.class).as("wa").where()
                    .critCondition(select(WorkActivityDetail.class).where().prop("workActivity").eq().prop("wa.id"), "comment", "waDetailsComments").and()
                    .critCondition(select(AffectedService.class).where().prop("workActivity").eq().prop("wa.id"), "service.key", "affectedServices").and()
                    .critCondition(select(Service.class).as("s").where().exists(SERVICE_MATRIX_SUBQUERY), "key", "relatedServices").and()
                    .critCondition(select(TeamManpower.class).where().prop("workActivity").eq().prop("wa.id"), "team.key", "manpowerTeams")
                    .yieldAll()
...
                    .modelAsEntity(ReWorkActivity.class);
```
All 4 uses of `.critCondition` in this case are combined with `.and()`, which cannot be changed at runtime. This is the main constraint when it comes to providing OR-grouping for collectional criteria. However, by defining all the necessary arguments not at the EQL model level, but at the level of corresponding criteria properties, we can support dynamic composition at runtime, which could take into account the OR-grouping information.

Each call includes 3 arguments:

1. A partial model of type `ICompoundCondition0<?>`. 

    For example, the first `.critCondition` call has partial model `select(WorkActivityDetail.class).where().prop("workActivity").eq().prop("wa.id")` as its argument.
2. A persistent property name of an entity used in that partial model, which needs to have selection criterial applied.
    
    In our example, property `comment` of entity `WorkActivityDetail` is specified.
3. A name for a collectional criteria, which is used to define search condition that should be applied to the property from step 2 in the context of a partial model in step1.
    
    In our example, collectional criteria  `waDetailsComments` is specified.

It is necessary to provide a way to specify arguments 1 and 2 at the level of collectional criteria (3), which is a `@CritOnly(MULTI)` property of the target synthetic entity.

Using our example, the current definition for `waDetailsComments` looks like this:

```Java
    @IsProperty
    @Title(value = "Comments ...",  desc = "...")
    @CritOnly(Type.MULTI)
    private String waDetailsComments;
```

Taking into account the limitations on arguments for Java annotations (i.e., only constant expression are permitted), the following represents the proposed approach for capturing arguments 1 and 2:

```Java
    @IsProperty
    @Title(value = "Comments ...",  desc = "...")
    @CritOnly(value = Type.MULTI,
              entityUnderCondition = WorkActivityDetail.class,
              propUnderCondition = "comment")
    private String waDetailsComments;
    protected final static ICompoundCondition0<WorkDetail> waDetailsComments_ = select(WorkActivityDetail.class).where().prop("workActivity").eq().prop("wa.id");
```

This definition should be processed in `DynamicQueryBuilder` to pass the captured arguments into dynamically constructed condition `.critCondition`, which would get combined with the rest of conditions using `.and()` be default and `.or()` in case of OR-grouping.

Hence, the implementation tasks:

- [x] Enhance annotation `CritOnly` with attributes `entityUnderCondition: Class<? extends AbsractEntity<?>>` and `propUnderCondition: String`. Strictly speaking `entityUnderCondition` is needed for validation only.
- [x] The OR-grouping mnemonic should only be enabled for `@CritOnly(MULTI)` with these extra attributes specified. In all other cases, it should still be restricted to current mnemonics "Not" and "Missing", which are supported as the EQL level when using `.critCondition`.
- [x] Enhance `CentreResource` and `tg-selection-criteria.js` to support marshalling of meta values for such collectional `@CritOnly` properties between the client and server. 
- [x] Ensure that mnemonics are persisted correctly as part of Entity Centre configurations.
- [x] Enhance `DynamicQueryBuilder` to apply `orGroup` for `.critCondition`s of `@CritOnly` collectional properties. This enhancement should rely on the fact that such collectional `@CritOnly` property has the associated partial query model defined as a static property with same name as the `@CritOnly` property + `_`. 

Here is a matrix that defines what mnemonics are and should be applicable to `@CritOnly` properties, depending on their type, partial-model and data type.

```
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| CritOnly Type  | Data Type    | Mnemonics without model (default)                 | Mnemonics with model                              | Mnemonics without model New with .critCondition |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| SINGLE         | Entity       |                                                   | Not, Missing, Or                                  | Not, Missing                                    |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| SINGLE         | String       |                                                   | Not, Missing, Or                                  | Not, Missing                                    |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| SINGLE         | Number/Money |                                                   | Not, Missing, Or                                  | Not, Missing                                    |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| SINGLE         | Date         |                                                   | Not, Missing, Or                                  | Not, Missing                                    |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| SINGLE         | Boolean      |                                                   | Or                                                |                                                 |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| MULTI          | Entity       | Not, Missing                                      | Not, Missing, Or                                  | Not, Missing                                    |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| MULTI          | String       | Not, Missing                                      | Not, Missing, Or                                  | Not, Missing                                    |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| RANGE          | Number/Money | Not, Missing, Exclude From/To                     | Not, Missing, Exclude From/To, Or                 | Not, Missing, Exclude From/To                   |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| RANGE          | Date         | Not, Missing, Exclude From/To, Prev/Curr/Next     | Not, Missing, Exclude From/To, Prev/Curr/Next, Or | Not, Missing, Exclude From/To, Prev/Curr/Next   |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
| RANGE          | Boolean      |                                                   | Or                                                |                                                 |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

### Expected outcome
Enriched search capabilities for Entity Centres, where collectional criteria could be used in conditional OR groups.

oleh-maikovych:
resolve #1818


jhou-pro (approved):

elvin-fms (approved):
Appears to be working:
1. Selection criteria:
image
2. Results
image

01es (approved):

PR #1851, opened 2022-07-13
Issue #1846 (improvements to Entity Centre concurrency)

### LINKED ISSUE #1846: Entity Centre: improve situation with rare self-concurrency conflicts
### Description

Over the last year or so there were several reports of low-level conflicting messages either reported as a server-side warnings or even propagated to the client-side as red toasts. The messages describe many different kinds of conflicts during `EntityCentreConfig` saving i.e. everything around entity centre running, saving, validation and other processes.

The above referenced entity centre processes indeed can be performed relatively simultaneously for the same user. Consider the example of where spinner (integer editor) is getting edited by using arrows in entity centre criteria. For some threshold, if we change the value quickly using arrow holding, simultaneous validation requests will be debounced / discarded by client-side logic. However, if spinner value is getting changed a little bit slower then it is possible to achieve two requests computing on server simultaneously. The other example is quick tapping on selection criteria checkbox; it is probably even easier to reproduce.

We can also do some quick actions on entity centre result-set that trigger simultaneous refreshes. The situation is getting compounded if, for example, there is SSE source that triggers additional refreshes. Also there can be multiple tabs open for the same centre and it can further increase the likelihood of simultaneous processes.

The main contention for simultaneous processes occurs during saving of `EntityCentreConfig` instances of different kinds. This entity represents single point of entity centre persistence. Due to the absence of requests / EQL queries queing (see `Asynchronous TG` in https://github.com/fieldenms/tg/projects?type=classic), we have chosen the following approach in 2018: if saving conflict occured then try to save again until successful. This approach leads to WARN logging on the server (mainly from `SessionInterceptor`), but always concludes saving successfully.

The problem is that this approach was not used consistently and everywhere. There are two different reasons for this:
1. "graceful repeating saving" is not possible in context of outer transaction, because conflicting exception would rollback everything
2. some processes were anticipated not to occur simultaneously, so "graceful repeating saving" was used only in necessary places

The main objective of this issue is to eliminate situations with red toasts about saving conflicts for the user. Also it would be nice to somehow reduce server-side warning cluttering (i.e., perhaps, warnings with `EntityCentreConfig` conflicts not to record stack traces).

### Expected outcome
Entity centre processes are always gracefully completed in pseudo-self-concurrent environment.

### Actual outcome 
Entity centre processes are completed with non-distracting low-level error messages about saving conflicts in some rare situations.

jhou-pro:
resolve #1846


platform-web-resources/src/main/java/ua/com/fielden/platform/web/resources/webui/CentreResourceUtils.java
```diff
@@ -262,6 +262,16 @@ public static boolean isRunning(final Map<String, Object> customObject) {
...
+    /**
+     * A predicate to determine whether {@code customObject} represents action with 'retrieveAll' option.
+     *
+     * @param customObject
+     * @return
+     */
+    public static boolean isRetrieveAll(final Map<String, Object> customObject) {
+        return Boolean.TRUE.equals(customObject.get("@@retrieveAll"));
+    }
+
```
jhou-pro, line 274: refs #1793

01es (approved):

Issue #1745, opened 2021-09-16
Activatable Entity: enhance ActivePropertyValidator to include processing of deactivatable dependencies


01es:
### Description

The current implementation for `ActivePropertyValidator`, which is executed when an activatable entity has its property `active` set to `false`, checks if that entity can be deactivated. The validation logic takes into account only references to the current entity. This is correct, but may not be exhaustive enough if the current entity has deactivatable dependencies, which are not counted towards dependencies for this entity. And as the result, the validator may permit deactivation only to report an error during saving of such change, when automatic deactivation for deactivatable dependencies is attempted.

Such behaviour impedes the user experience, where it is expected that all dependencies would be checked as part of the validation and all active dependencies would be presented in a single error message, rather then in portions -- one part during the validation and another during saving.

Therefore, it is required to enhance the validation scope for `ActivePropertyValidator` to also include all entities specified as deactivatable dependencies for the entity being deactivated (`@DeactivatableDependencies`), and report any active transitive dependencies in addition to the "direct" active dependencies. The algorithm should be recursive to take into account the possibility of deactivatable dependencies having deactivatable dependencies.

The actual error message should be constructed in such a way that transitive dependencies are clearly identified for the user to be able to find and handle those dependencies.

#### Implementation remark
Strictly speaking, the core change has to do with enhancing method `ActivePropertyValidator.getEntityDependantsMap` to collect entity types for deactivatable dependencies. Caching such information upon startup in the deployment mode could be useful.

Also, we currently maintain `ActivatableAbstractEntity.refCount` for performance reasons, which is not changed when dependencies are added or removed from the related deactivatable dependencies. This means that if `refCount == 0`, but there are deactivatable dependencies, the `ActivePropertyValidator` logic should still perform a full search. This is why we should consider whether the complexity to maintain `refCount` is worth the performance gains during entity deactivation. Potentially, it is worth refactoring the current approach and make `refCount` a function rather than a persistent property. The argument for this change is that entities do not get deactivated that often and that the actual computation of `refCount` would not be prohibitive from the performance perspective. Some of our real-world solutions should be used to ascertain this statement.

### Expected outcome
More complete validation during deactivation of entities.

01es:
When implementing this issue, please also consider https://github.com/fieldenms/tg/issues/1450, which could potentially be addressed at the same time. This is especially relevant if `refCount` would be removed in favor of the dynamic calculation of active references.

01es:
[...]

PR #1907, opened 2022-12-20
#1906 Implemented issue requirements.

### LINKED ISSUE #1906: Optimistic locking: automatic conflict resolution should be optional
### Description
By default, optimistic locking is applied to all persistent entities. Conflicts are recognised based on the versioning mechanism. And by default, automatic conflict resolution takes place, which may or may not succeed.

In practice, there are situations where automatic conflict resolution needs to be avoided, and instead a conflict error should be thrown, if the version-based mechanism identified a conflict.

All persistent entities have annotation `MapEntityTo`, which is a suitable place to control this behaviour. It is proposed to introduce attribute `autoConflictResolution` set to `true` by default, and inspect the value of this attribute in `PersistentEntitySaver.canResolveConflict` to return `false`, if `autoConflictResolution == false`, without attempting to resolve a source of conflict.

Entities, for which conflict recognition should strictly be based on `version`, should be annotated `MapEntityTo(autoConflictResolution = false)`.

### Expected outcome
Ability to control applicability of the automatic conflict resolution at the level of domain entities.

01es:
Resolve #1906


platform-dao/src/test/java/ua/com/fielden/platform/entity/AutomaticConflictResolutionTest.java
```diff
@@ -51,11 +53,10 @@ public void conflicting_changes_should_have_prevent_entity_saving() {
         } catch (final EntityCompanionException ex) {
             assertEquals(format("%s Tg Category [Cat1] could not be saved.", ERR_COULD_NOT_RESOLVE_CONFLICTING_CHANGES), ex.getMessage());
         }
-
     }
 
     @Test
-    public void concurrent_setting_of_activatable_property_should_not_lead_to_conflict_resolution_errors() {
+    public void concurrent_modification_of_refCount_due_to_new_activatable_dependencies_does_not_lead_to_conflict_resolution_errors() {
```
elvin-fms, line 59: Could this have an undesired side-effect on activatable entities with activatable dependencies if `autoConflictResolution` is turned off for them?
No PO lines in TG Air currently have activatable dependencies, but what if `autoConflictResolution` is turned off for some other entities, which do?

platform-dao/src/test/java/ua/com/fielden/platform/entity/AutomaticConflictResolutionTest.java
```diff
@@ -28,7 +30,7 @@ public void non_conflicting_changes_should_have_been_resolved() {
     }
 
     @Test
-    public void identical_conflicting_changes_should_have_been_resolved() {
+    public void identical_concurrent_changes_should_have_been_resolved() {
```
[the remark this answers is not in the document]
01es: So should we all!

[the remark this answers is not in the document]
01es: There was an attempt to update old test names, but this one must have slipped.

platform-dao/src/test/java/ua/com/fielden/platform/entity/AutomaticConflictResolutionTest.java
01es: It should not and that was verified. A test could be provided.

elvin-fms: Could be a good idea for complete certainty, unless #1745 is going to be implemented really soon.

platform-dao/src/test/java/ua/com/fielden/platform/entity/AutomaticConflictResolutionTest.java
```diff
@@ -64,15 +66,62 @@ public void concurrent_setting_of_activatable_property_should_not_lead_to_confli
         assertEquals(Integer.valueOf(3), co(TgCategory.class).findByKey("Cat1").getRefCount());
     }
 
+    @Test
+    public void concurrent_modification_of_refCount_due_to_new_activatable_dependencies_does_not_lead_to_conflict_resolution_errors_for_entities_with_not_auto_conflict_resolution() {
+        final TgBogieClass bc = co(TgBogieClass.class).findByKey("BC1");
+        assertEquals(Integer.valueOf(0), bc.getRefCount());
+
+        save(new_(TgBogie.class, "Bogie2").setBogieClass(bc));
+        save(new_(TgBogie.class, "Bogie3").setBogieClass(bc));
```
elvin-fms, line 75: Pardon me, but I don't really see concurrency here, it all looks very sequential.

I appreciate that all other tests here have two sequential invocations of `save()`, but this test is different. Other tests retrieve two instrumented instances of the same entity and then operate with them. In this test, I suppose `refCount` is not incremented on the non-instrumented instance of `bc` assigned to property `bogieClass`. Instead I suspect a fresh instrumented instance is retrieved during each sequential `save()`.

Actually, the same probably applies to `concurrent_modification_of_refCount_due_to_new_activatable_dependencies_does_not_lead_to_conflict_resolution_errors()`

elvin-fms (changes requested):

elvin-fms (approved):

PR #1989, opened 2023-05-31
Issue #1988

### LINKED ISSUE #1988: Attachments: sanitise values for `title` upon file upload + validation
### Description

In practice file names may contains leading, trainling, sequential whitespace as well as commas. File names are used as the default title for attachments. However, the default integrity constraints `RestrictExtraWhitespaceValidator` and `RestrictCommasValidator` prevent values that correspond to such filenames from being assigned to `Attachment.title`.

Instead of relaxing the default integrity constraints, the file names being assigned to `Attachment.title` should be sanitised by removing commas and extra spaces.

However, in case of hyperlink attachments, `Attachment.title` should admit valid URLs, which may include commas and should not be sanitised. Therefore, in this case values should be admitted as is. But in case of file attachments where users manually try to change `Attachment.title`, we should restrict extra spaces and commas per the default behaviour.

### Expected outcome
Ability to upload files with names that may contain leading, trainling, sequential whitespace as well as commas.

01es:
Resolve #1988

### DIFF
--- a/platform-pojo-bl/src/main/java/ua/com/fielden/platform/attachment/definers/AssignAttachmentTitle.java
+++ b/platform-pojo-bl/src/main/java/ua/com/fielden/platform/attachment/definers/AssignAttachmentTitle.java
@@ -21,7 +21,10 @@ public class AssignAttachmentTitle implements IAfterChangeEventHandler<String> {
     public void handle(final MetaProperty<String> property, final String value) {
         final Attachment attachment = property.getEntity();
         if (!attachment.isInitialising() && StringUtils.isEmpty(attachment.getTitle())) {
-            attachment.setTitle(value);
+            final String sanitizedValue = value.replaceAll(",", " ")    // replace commas with a space
+                                               .replaceAll("\\s+", " ") // replace sequential whitespace characters with a single space
+                                               .trim();                 // finally, remove leading and trailing whitespace characters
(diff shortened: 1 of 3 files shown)


platform-pojo-bl/src/main/java/ua/com/fielden/platform/attachment/definers/AssignAttachmentTitle.java
```diff
@@ -20,8 +19,11 @@ public class AssignAttachmentTitle implements IAfterChangeEventHandler<String> {
     @Override
     public void handle(final MetaProperty<String> property, final String value) {
         final Attachment attachment = property.getEntity();
-        if (!attachment.isInitialising() && StringUtils.isEmpty(attachment.getTitle())) {
-            attachment.setTitle(value);
+        if (!attachment.isInitialising() && isEmpty(attachment.getTitle())) {
+            final String sanitisedValue = value.replaceAll(",", " ")    // replace commas with a space
```
elvin-fms, line 23: If I understand correctly, a simple `replace()` will suffice. Or even `replace(',', ' ')`.

01es: @elvin-fms that is correct -- it would.

PR #2076, opened 2023-08-15
Issue #2055 - Entity Master: persist dimensions and position (locally)

### LINKED ISSUE #2055: Entity Master: persist dimensions and position (locally)
### Description

If changed by users, Entity Masters should persist their new dimensions and position into `localStore`.

This will enable users to adjust both thy dimensions and the position as required. This should be especially handy for tablets where  the behaviour of Entity Master is the same as on the desktop with large screens. For example, smaller Entity Masters such as those for locators, appear in the centre of the screen and with an onscreen keyboard visible, the list of autocompleted values is not visible, forcing users to drag the dialog every time they need to find something.

### Expected outcome
More customisability and usability improvement for tablets.

oleh-maikovych:
resolve #2055


jhou-pro (approved):

01es (changes requested):
As discussed in the Slack, improvements are required around the situation where dialogs appear outside of the browser window width.

01es (approved):

Issue #2084, opened 2023-08-25
Dialogs without prefferred dimensions gets callapsed into point with 0x0 dimension


oleh-maikovych:
### Description

Latest changes to `tg-custom-action-dialog` component bring another problem when dialogs without specified prefferred dimensions gets collapsed into 0x0 point. This happenes because when masters gets opened for the first time it tries to calculate it's prefferred dimensions if it doesn't exists. And this dimensions should be calculated before master gets opened and therefore doesn't have dimensions. Thats why calcualtion of prefferred dimensions returns 0x0. In order to solve that problem one should provide some default preffered dimensions or avoid prefferred dimension calcualtion before master gets opened.

### Expected outcome

Entity master without explicitely defined prefferred dimension should be visible.

### Actual outcome 

Entity master without prefferred dimensions gets callapsed into 0x0 point.

PR #2086, opened 2023-08-25
Issue #2084 (modern) - Dialogs without prefferred dimensions gets callapsed into point with 0x0 dimension

oleh-maikovych:
refs #2084


01es:
No specific changes pertaining to the Modern Java branch are required. So, this PR is not needed.

PR #2101, opened 2023-09-13
Issue #2092 - Persist dialog position and dimensions with screen resolution key

### LINKED ISSUE #2092: Persist dialog position and dimensions with screen resolution key
### Description

Issue #2055 enabled persistence of dialog position and dimensions locally. Which allowes user to open master there were it was opened last time. This works fine for system with only one display. But things are getting more sophisticated if user has more than one display with different resolution. In that case position and dimension saved on one displyed hardly match the other displayes. Therefor it is better to differentiate persisted positions and dimensions by screen resolution. It requires to include screen resolution into the key of localy saved postions and dimensions. Also it requires to react somehow on resolution changes.

### Expected outcome

Dialog position and dimension get saved with key that contains screen resolution and restores this values when screen resolution changes.

### Actual outcome 

Saved dialog position and dimension don't fit well into screen with different resolution.

oleh-maikovych:
resolve #2092


01es (changes requested):
Trying to maximise a dialog while data is being loaded has no effect. Afterwords, it becomes necessary to tap restore and maximise (i.e., twice tap the same action) to actually maximise the dialog. This behaviour is not observed on develop. The expectation is that tapping maximise would maximise dialogs even if the data is being loaded.

01es (approved):

PR #2152, opened 2024-01-22
Issue #2151- Decompose packaging of platform-annotation-processors

### LINKED ISSUE #2151: Decompose packaging of platform-annotation-processors
### Description
Currently, `platform-annotation-processors` module is packaged into a single *standalone* JAR with the
help of the Maven Shade plugin. *Standalone* means that all module dependencies are included in the
JAR. This is a requirement for annotation processors that may run in an isolated environment by
themselves (e.g., in an IDE). Hence *standalone*.

However, `platform-annotation-processors` is also used as a library - it provides annotations that
control annotation processors (e.g., `@SkipVerification`). This has undesired effects - usage of the
standalone JAR as a library breaks encapsulation. One consequence is that classes from
`platform-pojo-bl` are included in the classpath of TG applications twice, which creates ambiguity
of choice, but can be resolved by depending on the order of dependency declarations in `pom.xml` of
the dependent module (e.g., `*-pojo-bl` in a TG application).

To avoid breaking encapsulation of `platform-annotation-processors` it is proposed to package it
into 2 separate JARs:
1. Library JAR - standard packaging that includes the contents of `platform-annotation-processors` only.
2. Uber JAR - current practice.
To distinguish between them, Maven Classifier can be used.
Maven Shade plugin configuration will need to be modified.

See:
- https://maven.apache.org/plugins/maven-shade-plugin/shade-mojo.html (`shadedArtifactAttached`, `shadedClassifierName`)
- https://maven.apache.org/repositories/artifacts.html
- https://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html

### Expected outcome
`platform-annotation-processors` module is packaged via 2 jars: library and standalone.

homedirectory:
Resolve #2151

### DIFF
--- a/platform-annotation-processors/pom.xml
+++ b/platform-annotation-processors/pom.xml
@@ -90,6 +90,8 @@
                             <goal>shade</goal>
                         </goals>
                         <configuration>
+                            <shadedArtifactAttached>true</shadedArtifactAttached>
+                            <shadedClassifierName>standalone</shadedClassifierName>
                             <transformers>
                                 <!-- This transformer is needed to collect all annotation processors, including those that are defined in dependencies,
                                      and not end-up with the last processed file overriding our own. -->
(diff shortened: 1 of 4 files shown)


homedirectory:
The branch was created off `migration-to-java-11-with-EQL3`, thus will not be able to merge into `develop` without bringing everything else along. Should the branch be recreated off `develop`?

01es:
> The branch was created off `migration-to-java-11-with-EQL3`, thus will not be able to merge into `develop` without bringing everything else along. Should the branch be recreated off `develop`?

All good, branch `migration-to-java-11-with-EQL3` is still the main branch for modern Java.

01es (approved):

PR #2160, opened 2024-01-29
Issue #2159 - Entity Centre: self-concurrent auto-running for default configuration cause non-intrusive changes being lost

### LINKED ISSUE #2159: Entity Centre: self-concurrent auto-running for default configuration cause non-intrusive changes being lost
### Description

It has been discovered that dashboard-like Entity Centre with insertion points, containing embedded Entity Centre, may lose sorting configuration in that embedded Entity Centre.

The exact steps to reproduce this behaviour during normal application execution are not known. However, the same user can open multiple tabs/windows with such dashboard-like centre and change some entities in its embedded Entity Centre in different application instances. This will trigger simultaneous auto-running of the same autorunnable default configuration. And such simultaneous auto-running may cause non-intrusive changes being lost (sorting, column order/visibility, page capacity etc.).

Hypothetically, this problem may even be triggered on the same application instance. This is because both centres (the main and embedded) are being refreshed in case of entity changes or re-run in case of an SSE event. Running/refreshing of the main Entity Centre triggers a run/refresh of insertion points. And if those insertion points contain Entity Centres, they too get run/refreshed.

-------------------------------

Extra care has been previously taken to save different versions for Entity Centre configurations during running and other processes. In case of self-concurrent modifications, we use special `saveWithRetry` semantics to gracefully resolve conflicts by using the _last wins_ strategy.

However, clearing of criteria during auto-run was done in non-automatic way using the captureDiffs+removal+createNew+applyDiffs approach. This leads to a problem of the captureDiffs to be performed concurrently against already removed configuration (and thus non-intrusive changes get lost).

It is proposed to

 - [x] Make the above logic more atomic - fully avoid config removal
 - [x] Make running of the same Entity Centre by the same user sequential to avoid (or reduce) self-concurrency.

### Expected outcome
Non-intrusive changes always being preserved in the self-concurrent auto-running situations.

### Actual outcome 
Non-intrusive changes are intermittently being lost in the self-concurrent auto-running situations.

### Steps to reproduce
Concurrently save some entity in same embedded Entity Centre inside insertion point on different tabs/windows for the same user.

jhou-pro:
resolve #2159

### DIFF
--- a/platform-pojo-bl/src/main/java/ua/com/fielden/platform/entity/AbstractEntity.java
+++ b/platform-pojo-bl/src/main/java/ua/com/fielden/platform/entity/AbstractEntity.java
@@ -1495,7 +1495,7 @@ public abstract class AbstractEntity<K extends Comparable> implements Comparable
     }
 
     /**
-     * The main intent of this method is to support entity modification in rare situation while it it being marked as read-only.
+     * The main intent of this method is to support entity modification in rare situation while it is being marked as read-only.
      * Should be used with great care as it may alter the intended domain behaviour if used carelessly.
      * At this stage there is no reason for this setter to be used as part of the domain logic. */
     public void setIgnoreEditableState(final boolean ignoreEditableStateDuringSave) {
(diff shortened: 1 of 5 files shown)


01es (approved):
There are some aspects of the implementation of the approach for updating Entity Centre configuration that require further discussion and consideration. However, the issue can be considered as resolved in this specific narrow context.

PR #2200, opened 2024-02-28
Issue #2192 - Entity centres and masters: remove trailing and leading whitespace from search values in autocompleters

### LINKED ISSUE #2192: Entity centre: remove trailing and leading whitespace from entity-typed selection criteria
### Description

Leading and trailing whitespace is not permitted for key values. However, it is possible to enter such whitespace, usually unintentionally, when typing search strings into selection criteria for entity-typed properties (multi). The whitespace is then used as part of the search string, returning no results. And so, users need to realise that they have type a space before or after a search string, which is far from obvious.

To improve usability, it is proposed to automatically remove leading and trailing spaces from search strings used in entity-typed selection criteria when preparing a query.

### Expected outcome
Improved usability.

oleh-maikovych:
resolve #2192


01es:
Some of the implemented behaviour may need to be revisited. For example, removing trailing spaces where the intended search target is an entity description, may not be desirable.

PR #2220, opened 2024-03-27
Issue #2178

### LINKED ISSUE #2178: Formal grammar for EQL
### Description

It is proposed to describe a formal grammar for EQL to gain the following benefits:
* Facilitation of informal reasoning during development of the parser. As of now the only reference point is the fluent API itself, which is difficult to view as the grammar's representation since it is, in fact, its product.
* A better understanding of the state of EQL's fluent API can be gained by reverse engineering it into a formal grammar. 

The [Extended Backus-Naur form](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form) (EBNF) is a suitable notation candidate.

[Fling](https://github.com/OriRoth/fling) is a fluent API generator that provides means of expressing a grammar in EBNF using Java. Fling can be used to write down the grammar of EQL. The resulting grammar will be represented as a Java object, meaning that it can be transformed into arbitraty other formats if necessary.

### Expected outcome

A formal grammar for EQL expressed in a suitable notation.

homedirectory:
Resolve #2178 

Merge after #2209


homedirectory:
Became part of #2209

Issue #1891, opened 2022-11-03
EQL and Selection Criteria: support for meta-models


01es:
### Description

- [x] 1. `.critCondition()` API to support meta-models in order to avoid conversions to `String` using `.toPath`, like in the example below:
 ```Java
.critCondition(InventoryBin_.inventory().inventoryPart().key().toPath(), BatchInventoryTransfer_.partCrit().toPath())
```
- [x] 2. Interaction with EQL parameters needs to support meta-models. Currently params use `String` as keys. But also derived names for `@CritOnly` properties manipulate with `String` values.
      It would be great if something like the code below could be expressed in terms of meta-models:
      
```Java
final EnhancedCentreEntityQueryCriteria<?, ?> params = decompContext.selectionCrit();
final boolean showDuplicates = params.get(critName(RePmRoutineCalendar.class, "showDuplicates"));
```
### Expected outcome
Improved support for meta-models in TG API around EQL and selection criteria.

PR #2226, opened 2024-04-01
Issue #2223 - Reduce IConvertableToPath and String used for properties to a single common type

### LINKED ISSUE #2223: Reduce IConvertableToPath and String used for properties to a single common type
### Description

Paths of entity properties can be represented in 2 ways:

* via the traditional `String`
* via the compile-time safe `IConvertableToPath`

Due to the existence of 2 separate representations, it has been necessary to create 2 respective method signatures whenever the method needed to accept property paths as arguments.

It is proposed to eliminate this necessity by reducing these 2 types to a single common type that can represent property paths: `java.lang.CharSequence`.
Since `String` already implements `CharSequence`, this will only require `IConvertableToPath` to implement `CharSequence`.

This API change will also require recompilation of additional dependencies beside the TG applications, such as JMC.

### Expected outcome

Method signatures declaring parameter types `String` and `IConvertableToPath` that represent properties are reduced to a single signature declaring `CharSequence`.

homedirectory:
Resolve #2223, #1891

Merge after #2209


01es:
@homedirectory, does this really resolve https://github.com/fieldenms/tg/issues/1891?

homedirectory:
@01es, Yes, all requirements of #1891 are implemented:
1. `critCondition()`.
2. `CriteriaReflector.critName()`.

01es (approved):

Issue #2091, opened 2023-09-01
Reflector.isPropertyProxied needs to support property paths


01es:
### Description

Currently `Reflector.isPropertyProxied` can only check first-level properties. However, there are situations where this predicate is used in application to property paths (aka dot-notation expressions). For example, it is used in `AbstractEntity.get`, which accepts property paths.

The expectation is that if any property on the path is proxied, `true` should be returned.

This issue represents further improvement to the functionality developed as part of issues #89 and #415.

### Change overview

![image](https://github.com/user-attachments/assets/9b02c6bc-e18a-4f8a-ab0a-ff40e6d2a3bf)

### Expected outcome
Ability to recognised proxied properties in a property path using `Reflector.isPropertyProxied`.

### Notes

If #2223 is merged first, this issue should be implemented using `CharSequence`.

PR #2270, opened 2024-06-10
Issue #2258 - Typeful Metadata

### LINKED ISSUE #2258: Typeful Metadata
### Description

Completing the transition from EQL2 metadata to EQL3 metadata has posed some challenges due to
structural differences between the two. In particular, EQL3 metadata is less rich than its
predecessor, which requires its users to perform ad-hoc computations to retrieve the needed
information (e.g., to check whether a property represents a composite key), ultimately decreasing
the usefulness of the abstraction.

Historically, EQL metadata has been modelled with a simple abstraction -- a bag of attributes, i.e.,
a single Java class with multiple null-able fields. 
For example, metadata for entity properties has been modelled with a single class that has both
`column` and `expressionModel` fields with the former indicating a persistent nature and the latter
-- a calculated one. Since a property may have only one nature, only one of these fields may have a
non-null value in any given instance.

While such a simple abstraction may have been sufficient previously, the evolution of TG and its
modelling capabilities requires a more powerful abstraction to tackle the growing complexity.
Therefore, a *typeful* model for EQL metadata is proposed. Such model is based on the use of type
information, and provides stronger guarantees in the form of static properties of the program.
This model heavily relies on typechecking, which the Java compiler provides. Some of the
advantageous features supported by Java 21 are: sealed classes (can be used to model ADTs) and
pattern matching.

In a typeful model, types are used to represent domain concepts. In this case the primary concepts
are entities and their properties, which can be classified into several categories and modelled with
types accordingly (consult the [TG Wiki](https://github.com/fieldenms/tg/wiki/Property-definitions) for details).

### Expected outcome

Typeful metadata abstraction is introduced to assist with completing the migration to EQL3 metadata.

### Implementation

Inspiration can be drawn from the mirror-based API for Java annotation processing (`javax.lang.model`),
which has separated types into 2 levels: elements (`TypeElement`) and types (`TypeMirror`). The
same could be applied to TG metadata: elements -- entity and composite type definitions, types -- property types.
The latter (types) use the former (elements) and optionally include annotations for the purpose of
refinement. For example, numeric types can be refined by specifying their precision and scale.

homedirectory:
Resolve #2258, #2091.

This PR should be merged together with https://github.com/fieldenms/tglib-testing/pull/2.

## Summary of changes

There have been a number of breaking changes associated with the removal of EQL3 metadata and EQL2
metadata, therefore TG applications need to be adjusted accordingly.

Metadata:
* EQL3 and EQL2 metadata have been removed and their usages replaced by Typeful Metadata
* Stricter validation of property definitions during metadata generation: unsupported property types
  are detected early (in some cases an unsupported type is shown mercy and corresponding property is
  simply skipped).
* Changes to Property Metadata:
  * Properties with a composite or a union entity type are classified according to their nature in
    the same way as other properties. This hadn't been the case with EQL3 metadata, which ignored
    the nature of such properties, instead forcing the caller to look at their sub-properties for
    this information.

Other changes:
* Value range validation (from-to) is now supported for `BigDecimal`.
* All generated proxy entity types now implement `IProxyEntity`.
* Generated ID-only proxy entity types now implement `IIdOnlyProxyEntity`.
* Test entity types were modified by replacing unsupported property types with supported ones (e.g.,
`Double` by `BigDecimal`).

## Future work
* Encapsulate Hibernate type access in a standalone service. It should have the following API:
  
  ```java
  Optional<?> hibernateType(PropertyMetadata property)
  ```
* Encapsulate EqlTable access in a standalone service. It should have the following API:

  ```java
  EqlTable eqlTable(Class<? extends AbstractEntity<?>> entityType)
  ```
* Encapsulate TableStructForBatchInsertion access in a standalone service. It should have the following API:

  ```java
  TableStructForBatchInsertion tableStructForBatchInsertion(Class<? extends AbstractEntity<?>> entityType)
  ```
* Encapsulate database DDL generation in a standalone service. It should have the following API:

  ```java
  List<String> generateDatabaseDdl(Dialect dialect)
  ```
* Configure dependency injection and decouple `IDomainMetadata` from `QuerySourceInfoProvider`.


homedirectory:
Due to certain difficulties during merging, [1250408](https://github.com/fieldenms/tg/pull/2270/commits/1250408f7bf76b513e592ff7706c4d1446c10cca) was created by hand (with `git commit-tree`) and thus wasn't automatically signed.

01es (approved):

PR #2276, opened 2024-06-14
Issue #2271 - Entity Centre: insertion points to support non-default selection criteria for embedded centres upon run

### LINKED ISSUE #2271: Entity Centre: insertion points to support non-default selection criteria for embedded centres upon run
### Description

Currently, when running Entity Centres, all insertion points with embedded Entity Centres, cances selection criteria values that might have been entered by the user. This behaviour ensures that all subordinate queries, run by such embedded Entity Centres, are aligned with the main Entity Centre and its selection criteria. Otherwise, it is very likely to create a situation where insertion points display an empty result or a subset of the data that does not fully align with the main result.

However, there are situations, such as in dashboards, which extensively utilise insertion points, there is no single "main" result. In other words, the embedded into the insertion points Entity Centres, are independent. In such situations running happens as the means for refreshing the data in those independent insertion points, which currently clears any selection criteria that users might have entered. This situation needs to be improved.

It is proposed to introduce a configuration parameter for insertion points (or embedded entity centres more generally) to:

- [x] 1. Skip clearing selection criteria for embedded Entity Centres during run.

- [x] 2. Keep the currently loaded save-as configuration, which should retain any unsaved modifications upon run.

In addition:

- [x] 3. Provide a visual cue by highlighting the filter action in blue (`--paper-light-blue-600`) in cases where changes to selection criteria were made, but not saved. This is to bring the user's attention to the fact that custom selection criteria are used. Such visual cue would be applied generically to embedded and standalone sentres, irrespective of whether it is configured to run automatically or not.

- [x] 4. Improve indication of the loaded save-as configurations for insertion points by showing the configuration title in the insertion point's title bar (in `()`) - not the main application title bar as it happens for standalone centres.

### Change overview

This issue introduces parameterisation for Entity Centre configuration `.runAutomatically`. A single possible value for this parameter is `RunAutomaticallyOptions.ALLOW_CUSTOMISED`. If specified for either standalone or embedded centre, any selection criteria changes would get preserved upon auto-running. 

Also, this parameter preserves any loaded save-as configurations upon auto-run. For example, if a user loads some save-as configuration for an embedded centre, that configuration would be used for auto-run until such time as user changes it (loads the default or any other save-as configuration).

Without this parameter the default behaviour is applied. More specifically the default configuration is always loaded before auto-run is performed.

### Expected outcome
Ability to configure Entity Centres, embedded into insertion points, not to clear selection criteria when running.

jhou-pro:
resolve #2271


oleh-maikovych (approved):

01es (changes requested):
Additional changes are required as per updated issue https://github.com/fieldenms/tg/issues/2271

01es (approved):

PR #2314, opened 2024-09-13
Issues #2271 #2275 -- Entity Centre: insertion points to support non-default selection criteria for embedded centres upon run + Entity Centre: floating and D'n'D for Insertion Points

### LINKED ISSUE #2271: Entity Centre: insertion points to support non-default selection criteria for embedded centres upon run

### LINKED ISSUE #2275: Entity Centre: floating and D'n'D for Insertion Points
### Description

Insertion Points (IPs) have their designated placement as part of the overall Entity Centre layout. They can be resized vertically and expanded to full screen.

As the next step in the evolution of IPs, it is required to support:

- [ ] 1. Ability to float an Insertion Point by unsnapping it. Floated IP should behave identical to an Entity Master - be resizable, minimisable, movable, expendable. The dimensions of a floated IP should be the same as it was before floating, but its position should be centred.

  A floatable IP should remain associated with its Entity Centre (i.e., be visible only when its centre is in view) and it should be possible to snap the IP back to its last position (let's call this "snap position"), having the height set to the same value it had before being floated (unsnapped).
  
  Additional title bar buttons are required to enable floating and snapping IP back to its last snap position.

- [ ] 2. Ability to minimize/restore an Insertion Point while snapped. This would be analogous to how minimise/restore works for Entity Master, but the IP would remain snapped in place and restoring would revert only the height. IPs below the minimised IP would move up, and down when restored.

- [ ] 3. Ability to change a snap position for IPs. Users should be able to D'n'D a snapped IP by its title bar above or below of any other IP. The width of the dropped IP should attain the width of the target section, while the height should remain, potentially pushing IPs below further to the bottom.

    Once dropped, the new snap position should become the last snap position for that IP.

- [ ] 4. Restoration of dimensions associated with double-tapping the bottom right corner should restore the dimensions, and if in floated state, the position (i.e., centered). There is no need to snap a floated IP back to its last snap position.

All manipulations with IPs should be stored locally in order to be able to restore the state after reloading the application in the browser.

### Expected outcome
Improved control by users over the size and position of Insertion Points.

jhou-pro:
resolve #2271 resolve #2275


jhou-pro:
Use original #2298 pull request.

PR #2324, opened 2024-10-07
Issue #185 - WYSIWYG editor for properties of type RichText

### LINKED ISSUE #185: WYSIWYG editor for properties of type RichText
### Pre-requsites

Issue #348 needs to be implemented first. However, basic research of the subject of developing a component for HTML editing and rendering can proceed independently.

### Description

It would be great to support some markup language (e.g. HTML) as part of a multiline text editor. However, such support is not trivial due to the need for separation of formatting from the context upon persisting an associated string property into a database. Such separation is required to facilitate search, which would be impeded by presence of formatting tags in the value of string properties.

Property type `RichText` should be the target type for this editor.

##### Implementation details

Investigate if [StackEdit](https://github.com/benweet/stackedit) can be used. It is highly desirable to have a good WYSIWYG editor.

If not, investigate a possible use of https://github.com/markedjs/marked for implementing markdown rendering.

Also [ToastUI-editor](https://ui.toast.com/tui-editor) is a good alternative of HTML or Markdown WYSIWYG editor. Investigate how it can be integrated with shadow-dom

Special care should be taken for text sanitisation (e.g., https://github.com/cure53/DOMPurify).

### Change overview

A rich text editor on an entity master can be created by calling method `asRichText()` after `addProp()`. This method invocation creates a flexible rich text editor that automatically adjusts its size when the master is resized. This means that the height of the rich text editor cannot be changed manually by the user; it only changes when the entity master is resized.

Additionally, it is important to adjust the layout to ensure that the cell containing the rich text editor is **flexible** in the vertical direction. One may also specify a minimum height for the flexible rich text editor. This minimum height prevents the editor from shrinking below the desired height when resizing the entity master, and the default minimum height is 25 pixels.

The minimum height can be set by calling the `withMinHeight(minHeight)` method after invoking `asRichText()`. The `minHeight` parameter is an integer representing the minimum height of the rich text editor in pixels. The following example demonstrates how to create a flexible rich text editor with a minimum height of 100 pixels:

```java
.addProp("richDesc").asRichText().withMinHeight(100)
```

To create a rich text editor with a fixed height, call `withHeight(height)` after `asRichText()`, where `height` is an integer representing the height in pixels. A rich text editor with a fixed height allows users to manually adjust its height by dragging the resize icon in the bottom-right corner of the editor. The following example creates a rich text editor with a fixed height of 350 pixels:

```java
.addProp("richDesc").asRichText().withHeight(350)
```

It is important to ensure that, in this case, the layout for the cell containing the rich text editor is not flexible in the vertical direction. Otherwise, it may lead to undesirable layout behavior.

Additionally, a minimum height for a rich text editor with a fixed height can be specified by calling `setMinHeight(minHeight)`, where `minHeight` is an integer representing the minimum height of the rich text editor in pixels. This prevents users from shrinking the editor below the specified minimum height. The following example creates a rich text editor with a fixed height of 350 pixels and a minimum height of 100 pixels:

```java
.addProp("richDesc").asRichText().withHeight(350).withMinHeight(100)
```

### Expected outcome

Support for wiki-like text with rich text rendering as a property editor for properties of type `RichText`.

jhou-pro:
resolve #185


01es:
A new PR should be created.

PR #2812, opened 2026-08-07
Issue #2434 - Application update: notification with prompt

### LINKED ISSUE #2434: Application update: immediate top bar notification
### Description

It has been discussed and agreed that users would benefit of knowing if (when?) the application update has occurred. This is especially important in multi-tab/window usage scenario where the user may have refreshed the current window / tab, but still forgotten about others, which would lead to confusing errors.

It is proposed to immediately notify all browser tabs / windows once SSE connection was re-established and application version is different from the one on the client.

- [ ] Consider spreading out SSE notifications through a 1 minute interval as per newly established practise to reduce initial server load.
- [ ] Use `tg-message-panel` top bar for this purpose; ensure ability to close the notification if user wants to.
- [ ] Use generated `<title>@title</title>` in main `index.html` resource for comparing...
  or alternatively
- [ ] ... provide `application version` as a first-class concept and use it for title generation too (slightly more involved).

refs #1072

### Expected outcome

Automatic encouragement for users to update their client applications in all tabs / windows as soon as a new version is available.

jhou-pro:
Resolve #2434 

# To be completed by the pull request creator

This section should be completed with reference to section [Preparing PR](https://github.com/fieldenms/devops/wiki/Code-and-PR-reviews#preparing-pr) of the [Code and PR reviews](https://github.com/fieldenms/devops/wiki/Code-and-PR-reviews) wiki page.

- [x] Create the pull request as a draft by tapping the dropdown arrow on the 'Create pull request' button under the pull request description (below the text box where this description is being edited) and changing the default `Create pull request` to `Draft pull request`.
      Or, if the pull request has already been created, convert it to draft by tapping the "Convert to draft" link beneath the "Reviewers" section.

- [x] A self-review of all changes has been completed, and the changes are in sync with the issue requirements.

- [x] Changes to the requirements have been reflected in the issue description.

- [x] Any "leftovers" such as sysouts, printing of stack traces, and any other "temporary" code, have been removed.

- [x] Minor refactorings, such as renamings, extraction of constants, etc., have been addressed.

- [x] Developer documentation (e.g., comments, Javadoc), have been provided where required.

- [x] All existing and new Java tests pass successfully by running them with Maven.

- [x] All existing and new Web tests pass successfully.

- [x] Changes have been inspected for possible NPE situations, and the changes are sufficiently defensive.

- [x] The correct base branch has been selected for these changes to be merged into.

- [x] The latest changes from the base branch have already been merged into this feature branch (and tested).

- [x] Added a change overview to the issue description or as a wiki page, referenced in the issue description.
      Some issues might be very descriptive and serve in place of a wiki page.
      In such cases consider adding label `Wiki like` to the issue.

- [x] Changes subject to performance considerations have been evaluated, and tested against production-size data if applicable.

- [x] This pull request does contain significant changes, the section "Significant changes" below is completed and at least one Senior Software Engineer with the relevant area of expertise has been selected as reviewer.

- [x] The `In progress` label has been removed from the issue.

- [x] The `Pull request` label has been added to the issue.

- [x] The pull request has been made ready for review by tapping the "Ready for review" button below the list of commits on the pull request page.

## Significant changes

This pull request contains significant changes as defined in [the wiki page](https://github.com/fieldenms/devops/wiki/Code-and-PR-reviews#significant-changes).

Details are as follows:

The new toast component with distinct UX from existing ones has been added. The details of how it should look has been agreed with @01es.
- [ ] First line of text should be white on dark background and be bold to emphasise it.
- [ ] Second line of text is less important and may explain in detail the main message: grey and thin font; wraps if long.
- [ ] Buttons can be custom, but for app update notification the main `Reload` button should be raised with blue background, with white and bold text caption. And no uppercasing.

# To be completed by the pull request reviewer

This section should be completed with reference to section [Performing PR review](https://github.com/fieldenms/devops/wiki/Code-and-PR-reviews#performing-pr-review) of the [Code and PR reviews](https://github.com/fieldenms/devops/wiki/Code-and-PR-reviews) wiki page.

- [x] The `In progress` label has been added to the pull request in GitHub.

- [x] The issue requirements have been read and understood (along with any relevant emails and/or Slack messages).

- [x] The correct base branch is specified, and that base branch is up-to-date in the local source.

- [x] The issue branch has been checked out locally, and had the base branch merged into it.

- [ ] All automated tests pass successfully.

- [ ] Ensure the implementation satisfies the functional requirements.

- [ ] Ensure that code changes are secure and align with the established coding practices, including code formatting and naming conventions.

- [ ] Ensure that code changes are documented and covered with automated tests as applicable.

- [ ] Ensure that code changes are well-suited for informal reasoning.

- [ ] Ensure that changes are documented for the end-user (a software engineer in the case of TG, or an application user in the case of TG-based applications).

- [ ] If there are significant changes (described above), special attention has been paid to them.
      Marked the task items in section "Significant changes" as completed to indicate that corresponding changes have been reviewed, improved if necessary, and approved.

- [ ] The issue or issues addressed by the pull request are associated with the relevant release milestone.

# To be completed by the pull request reviewer once the changes have been reviewed and accepted

- [ ] The changes have been merged into the base branch (unless there is a specific request not to do so, e.g., they are to be released to SIT).

- [ ] The issue branch has been deleted (unless the changes have not been merged - see above, or there is a specific request not to do so).

- [ ] The `In progress` label has been removed from the pull request.

- [ ] The `Pull request` label has been removed from the issue.


01es:
The design is sound: announce the version on SSE (re)connect, compare against the version the client booted with, prompt in a toast that survives until dismissed.
The findings below are ordered by consequence.

---

## Blockers

### A. `TgTooltipBehavior` on `tg-message-panel` inflates the login bundle 3.5×

`platform-web-ui/src/main/web/ua/com/fielden/platform/web/components/tg-message-panel.js:132`

Measured on this branch:

| | before | after |
|---|---|---|
| `login-startup-resources-vulcanized.js` | 338,038 B | 1,189,327 B |
| `ProseMirror` occurrences | 0 | 53 |
| `toastui` occurrences | 0 | 448 |
| `DOMPurify` occurrences | 0 | 8 |

The import chain:

```
tg-message-panel.js:132   behaviors: [ TgTooltipBehavior ]
  → tg-tooltip-behavior.js:2      import '/resources/components/tg-tooltip.js'
    → tg-tooltip.js:5             import { tgRichTextStyles } from '.../tg-rich-text-styles.js'
      → tg-rich-text-styles.js:4  import { toastuiEditorStyles } from '.../toastui-editor-lib.js'
```

`toastui-editor-lib.js` lines 5–7 re-export the editor and DOMPurify explicitly:

```js
export { default as purify } from '../dompurify/dist/purify.es.mjs.js';
import { Editor as ToastUIEditor } from '../@toast-ui/editor/dist/esm/index.js';
export { Editor as default } from '../@toast-ui/editor/dist/esm/index.js';
```

so consuming only `toastuiEditorStyles` still drags the whole editor into the graph.

The +851 KB lands on `login.html`, `logout.html`, `error-page-template.html` and the four
password-reset pages — the unauthenticated entry points, where the service-worker cache is
coldest and first-load latency matters most — in order to render one `tooltip-text` on a
Close button.

It is also redundant in the main application: `tg-app-template.js:260` already carries
`TgTooltipBehavior` and hosts `<tg-message-panel>` at line 57, so the panel's tooltip already
worked there. The fix was only ever needed on the login path, which is exactly where it is
most expensive.

**Fix.** Break `tg-tooltip.js`'s dependency on the rich-text styles, or split the editor
styles off `toastui-editor-lib.js`, rather than paying the editor's weight to show a tooltip.
Consider splitting this out as its own issue.

### B. A throwing `appVersion()` override takes down SSE for every client

`platform-web-resources/src/main/java/ua/com/fielden/platform/web/sse/EventSourceDispatchingEmitter.java:128`

```java
final var appVersion = appVersionSupplier.get();   // outside the try
if (!isBlank(appVersion)) {
    try { emitter.event(APP_VERSION_EVENT_NAME, appVersion); }
    catch (final IOException ex) { ... }
}
```

`IWebUiConfig.appVersion()` is a documented extension point, and the issue's *Future Work*
says applications will eventually supply a real version — likely read from a manifest,
a property file, or a DB row, i.e. something that *can* throw. That exception propagates out
of `registerEmitter` into `SseServlet.doGet:193`, which deregisters the emitter and rethrows
as `ServletException`. The client then retries every 15 s and fails every time, so the whole
eventing subsystem dies for a cosmetic notification. The constructor at line 69 does not
`requireNonNull` the supplier either, so a `null` has the same effect.

**Fix.** Move `appVersionSupplier.get()` inside the guarded block and catch `Exception`
rather than only `IOException`; add a null check in the constructor.

---

## Correctness

### C. The two sticky messages silently overwrite each other

`tg-sticky-toast.js:74`, `tg-app-template.js:920`, `tg-entity-centre.js:1059`

Both the component doc (*"never gets overridden by, and never overrides, other messages"*) and
the PR description make this claim, but it only holds against *transient* toasts. There is
exactly one `stickyToastElement` (`tg-sticky-toast.js:164`) and `showMessage` replaces
`messageText` / `messageDetail` / `messageActions` and `_messageHandlers` wholesale.

The two callers fire in the same circumstance — right after a deployment:

1. Tab reconnects → *"A new application version is available."*
2. User opens a centre whose insertion-point DSL changed in that same deployment →
   `_restoreInsertionPointOrder` → *"Custom layout was reset."* replaces it.
3. User closes that → the update prompt is gone permanently, because `_notifiedAppVersion`
   was already set at `tg-app-template.js:920` and is never re-shown for that version.

That is precisely the stale-tab scenario the issue set out to prevent.

**Fix.** Cheapest: set `_notifiedAppVersion` when the user actually dismisses (in the
`later` / `reload` handlers) rather than at display time. A small queue in `tg-sticky-toast`
would fix it generically.

### D. The `IOException` comment describes something that cannot happen

`EventSourceDispatchingEmitter.java:133`

`EventSourceEmitter.event(...)` (`EventSourceEmitter.java:79-91`) catches `IOException`
internally and calls `close()`; it never propagates. So the catch is unreachable for the
production emitter, and the real failure mode — the emitter closing and deregistering itself
via its close callback — is neither logged here nor "non-critical". Worse, `registerEmitter`
still returns `successful(emitter)` and `SseServlet` logs *"SSE subscription … completed"*
for a connection that is already dead.

Keep the catch (the interface declares the exception), but make the comment match reality.

### E. `checksums` is a better default than `title()`

`platform-web-ui/src/main/java/ua/com/fielden/platform/web/app/IWebUiConfig.java:238`

The issue defers first-class versioning to *Future Work*, so `title()` as an interim proxy is
a deliberate choice. But `AbstractWebUiConfig` already holds
`SequencedMap<String, String> checksums` (line 131, loaded at line 185) — hashes of the
vulcanised resources, which change exactly when a new client build ships. That is a real
deployment identity, already present, and it would avoid two problems with the current
default:

- **Inert by default.** The platform's own `WebUiConfig` passes the constant
  `"TG Test and Demo Application"` (`WebUiConfig.java:134`), so for any application that does
  not override `appVersion()` the whole feature — SSE event, client listener, sticky toast —
  is silently dead, and nothing surfaces that.
- **False positives.** Renaming a title for cosmetic reasons (*"Acme Prod"* → *"Acme
  Production"*) prompts every connected client to reload with no deployment at all.

### F. `isStale` configuration and the update prompt collide

`tg-app-template.js:841` already raises a `Configuration load error` toast when the
configuration came from the `localStorage` fallback (`tg-app-resource-loader.js:31-58`). In
that same state `bootAppVersion` is a value captured in an earlier session, so the version
prompt fires too and the user sees two conflicting messages.

This is a UX conflict rather than a defect — a client running SW-cached resources with a
stale configuration genuinely *may* be out of date, so the prompt is not necessarily wrong.
Worth deciding deliberately which message wins.

---

## Conventions

### G. `isBlank` imported from Tika

`EventSourceDispatchingEmitter.java:21`

```java
import static org.apache.tika.utils.StringUtils.isBlank;
```

Tika's `StringUtils` is an internal helper of a document-parsing library, present only
transitively via `platform-pojo-bl` for attachment MIME detection; `platform-web-resources`
does not declare Tika at all. The codebase convention is `org.apache.commons.lang3.StringUtils`
(192 references), and commons-lang3 is already imported in this very file at line 3. Looks
like an IDE auto-import.

The same slip already exists at `EntityMasterUrlProvider.java:21` — pre-existing, but these
are the only two places in the codebase that reach into Tika's internal utils, so worth
sweeping both.

### H. Eager log formatting in new code

`EventSourceDispatchingEmitter.java:134`

CLAUDE.md requires the lazy form:

```java
LOGGER.warn(() -> "Could not announce application version [%s] to a newly connected SSE client.".formatted(appVersion), ex);
```

The surrounding `format(...)` calls are pre-existing, but this is new code, and the same
commit correctly removed two pointless `format(...)` wrappers at lines 229 and 240.

### I. `MutableBoolean` is a first for this codebase

`EventSourceDispatchingEmitter.java:3,110`

No other file uses `org.apache.commons.lang3.mutable`. `AtomicBoolean` is already imported at
line 17 for `isActive` and does the same job. Not wrong — just an avoidable addition to the
vocabulary.

---

## Minor

- **`paper-button` not imported where it is styled** — `tg-sticky-toast.js:1`. The module
  declares the `.sticky-toast .action` contract for `paper-button` (lines 40–46) and
  advertises actions in `showMessage`'s doc, but never imports `paper-button.js`. It works
  today only because both callers import it independently. Any third caller gets an
  un-upgraded inert element. Add the import here and drop the incidental one from
  `tg-app-template.js:11`.
- **`innerHTML` sink** — `tg-sticky-toast.js:112-114` assigns raw HTML for `text`, `detail`
  and `actions`, and `tg-app-template.js:922` interpolates the SSE-supplied version into it.
  Server-controlled today, so not an attack path, but `showStickyToast` is exported for use
  "from anywhere in the application"; escape by default and require an explicit opt-in for
  markup.
- **Announcement can be dropped on first connect** — `tg-event-source.js` is the first import
  in `startup-resources-origin.js:3` and opens the connection at module-evaluation time,
  before `addAppTemplate()` creates `tg-app-template` and its `ready` attaches the listener
  (`tg-app-template.js:877`). The server announces only for a new emitter
  (`EventSourceDispatchingEmitter.java:114-116`) and never re-announces, so a slow client can
  lose it. Benign in the normal path — the boot version comes from a fresh
  `/app/configuration` fetch and matches — but it is the `isStale` path (see F) where it
  would matter. Retain the last announced version in `tg-event-source.js` for
  `tg-app-template` to read on `ready`.
- **Listener cannot be removed** — same line; the arrow function is not stored, so `detached`
  (line 910) cannot remove it the way it removes `beforeunload`.
- **`hide()` leaves the message live** — `tg-sticky-toast.js:132` closes the overlay but keeps
  the innerHTML and `_messageHandlers`, so the dismissed message's handler closures stay
  reachable. Clear both on close.
- **Accessibility** — `TgPaperToast._openedChanged` fires `iron-announce` with `this.text`
  (`tg-paper-toast.js:29-31`), which is empty for the sticky toast because all content is
  slotted. Screen-reader users get no notice of either message. Set
  `this.$.stickyToast.text` to the plain-text form before `show()`.
- **Action styling is copy-pasted** — `tg-sticky-toast.js:40` declares a `.action` default
  that both callers immediately override inline (`tg-app-template.js:930`,
  `tg-entity-centre.js:1062`), with the secondary-button CSS duplicated verbatim. Declare
  `.action.primary` / `.action.secondary` in the component and let callers pass a class.
- **Dead `setAttribute`** — `tg-sticky-toast.js:59` sets `style` on the `<template>`, but only
  `.content` is appended, so it never applies. Copied verbatim from
  `tg-toast-behavior.js:48`, where it is equally inert — an existing idiom rather than a new
  mistake.
- **`paper-styles/color.js`** — `tg-app-template.js:12` is already pulled in by
  `tg-sticky-toast.js:2`, so the "Both are required" comment overstates it. Harmless.
- **Comment typo** — `tg-event-source.js:59`: "through a window event`.,` so that".
- **Binary compatibility** — the implicit no-arg `EventSourceDispatchingEmitter()` constructor
  is gone. One in-platform call site, but it is a public platform class; worth a release note.
- **No tests.** `EventSourceDispatchingEmitter` has no test class at all, and the new logic is
  easy to pin with a fake `IEventSourceEmitter`: announced once for a new key, *not*
  re-announced for an existing key, skipped when blank, and — for B — registration still
  succeeds when the supplier throws.
- **Worth eyeballing** — `TgToastBehavior.ready` forces `#label { flex: 1 }` and
  `tg-paper-toast` is `--layout-horizontal`. With the sticky toast's label empty, spare width
  goes to `#label` and pushes the message right. Visible only when a wider transient toast
  stacks above and stretches the container — precisely the co-existence case this component
  exists for.

---

## Checked and sound

- Announcing **outside** `computeIfAbsent`'s mapping function is the right call and worth the
  `MutableBoolean` dance — a failed `event()` closes the emitter, whose close callback calls
  `deregisterEmitter` → `register.remove`, which inside the mapping function would be a
  recursive map update.
- `@apply` inside the document-level `<custom-style>` does work — apply-shim registers a
  `transformCallback` with the custom-style interface, and `tg-toast-behavior.js` already
  depends on the same mechanism.
- The sticky toast is `appendChild`-ed while transient toasts `prepend`, so it genuinely stays
  at the bottom of `#toastContainer`.
- `duration="0"` disables auto-close in `TgPaperToast._openedChanged`.
- `.sticky-toast-actions[hidden]` is needed and correctly specific — the layout mixin does set
  `display`.
- `window.location.reload()` is sufficient: the service worker revalidates every static
  resource against the server checksum on each fetch.
- `--tg-main-pannel-color` matches the (misspelled) property actually set at
  `tg-app-resource-loader.js:176`.
- No SSE event-name collision with `connection` / `exception` / `message` / `completed`.
- `ApplicationConfigurationResourceTest` asserts individual keys, not the key set, so the new
  `appVersion` entry does not break it.
- `checksums.json` is updated for both vulcanised bundles.
- Nested `TgTooltipBehavior` (panel inside app template) is an established pattern here —
  `tg-entity-grid-inspector`, `tg-entity-centre-insertion-point` and others already nest
  inside `tg-app-template`.

---

## Bottom line

Two blockers: **A** (851 KB regression on the unauthenticated path) and **B** (a throwing
`appVersion()` override takes down SSE for every client). **C** is the main functional gap,
since it can lose the very notification this issue exists to deliver.

01es (changes requested):
Refer to the extended comment above.

